PR #11, opened 2024-03-12
feat: minimum componentize changes

guybedford:
This is the bare minimum changes I need to get ComponentizeJS to work on top of StarlingMonkey.

* `cabi_realloc` as a weak symbol so it can be overridden to support the bulk deallocation strategy we use
* A new `script_value()` on the engine, allowing to get the value of the top-level script. For a script it would be the last statement value, and for a module it is defined to be the module namespace itself.
* Allowing the console `console_builtin_log` to be overridable as a weak symbol
* Supporting cycles in the script loader (since ComponentizeJS has a cycle with its bindings module)
* Removing the full path in the debug logs for script errors, showing only the base-path-relative filename
* Fixing relative resolution to be parent-relative always
* Adding support for `RelWithDebInfo` build target

I kept it to the bare minimum to avoid too much design discussion and just get a prototype together.

Feedback welcome.

### DIFF
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -54,7 +54,7 @@ public:
    * Get the JS value associated with the top-level script execution -
    * the last expression for a script, or the module namespace for a module.
    */
-  JS::PersistentRootedObject script_value();
+  HandleObject script_value();
 
   bool has_pending_async_tasks();
   void queue_async_task(AsyncTask *task);
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -50,6 +50,12 @@ public:
 
   bool run_event_loop(MutableHandleValue result);
 
+  /**
+   * Get the JS value associated with the top-level script execution -
+   * the last expression for a script, or the module namespace for a module.
+   */
+  JS::PersistentRootedObject script_value();
+
   bool has_pending_async_tasks();
   void queue_async_task(AsyncTask *task);
   bool cancel_async_task(int32_t id);
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -54,7 +54,7 @@ public:
    * Get the JS value associated with the top-level script execution -
    * the last expression for a script, or the module namespace for a module.
    */
-  HandleObject script_value();
+  HandleValue script_value();
 
   bool has_pending_async_tasks();
   void queue_async_task(AsyncTask *task);
--- a/runtime/engine.cpp
+++ b/runtime/engine.cpp
@@ -343,7 +343,7 @@ void api::Engine::enable_module_mode(bool enable) {
 }
 
 JS::PersistentRootedObject SCRIPT_VALUE;
-JS::PersistentRootedObject api::Engine::script_value() {
+HandleObject api::Engine::script_value() {
   return SCRIPT_VALUE;
 }
 
--- a/runtime/engine.cpp
+++ b/runtime/engine.cpp
@@ -342,8 +342,8 @@ void api::Engine::enable_module_mode(bool enable) {
   scriptLoader->enable_module_mode(enable);
 }
 
-JS::PersistentRootedObject SCRIPT_VALUE;
-HandleObject api::Engine::script_value() {
+JS::PersistentRootedValue SCRIPT_VALUE;
+HandleValue api::Engine::script_value() {
   return SCRIPT_VALUE;
 }
 
@@ -351,12 +351,12 @@ void api::Engine::abort(const char *reason) { ::abort(CONTEXT, reason); }
 
 bool api::Engine::eval_toplevel(const char *path, MutableHandleValue result) {
   JSContext *cx = CONTEXT;
-  if (!scriptLoader->load_top_level_script(path, result)) {
+  RootedValue ns(cx);
+  if (!scriptLoader->load_top_level_script(path, &ns)) {
     return false;
   }
-  if (result.isObject()) {
-    SCRIPT_VALUE.init(cx, JS::RootedObject(cx, &result.toObject()));
-  }
+
+  SCRIPT_VALUE.init(cx, ns);
 
   // Ensure that any pending promise reactions are run before taking the
   // snapshot.
--- a/runtime/engine.cpp
+++ b/runtime/engine.cpp
@@ -342,12 +342,20 @@ void api::Engine::enable_module_mode(bool enable) {
   scriptLoader->enable_module_mode(enable);
 }
 
+JS::PersistentRootedObject SCRIPT_VALUE;
+JS::PersistentRootedObject api::Engine::script_value() {
+  return SCRIPT_VALUE;
+}
+
 void api::Engine::abort(const char *reason) { ::abort(CONTEXT, reason); }
 
 bool api::Engine::eval_toplevel(const char *path, MutableHandleValue result) {
   JSContext *cx = CONTEXT;
   if (!scriptLoader->load_top_level_script(path, result)) {
     return false;
+  }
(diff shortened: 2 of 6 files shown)


include/extension-api.h
```diff
@@ -50,6 +50,12 @@ class Engine {
 
   bool run_event_loop(MutableHandleValue result);
 
+  /**
+   * Get the JS value associated with the top-level script execution -
+   * the last expression for a script, or the module namespace for a module.
+   */
+  JS::PersistentRootedObject script_value();
```
tschneidereit, line 57:
```suggestion
  HandleObject script_value();
```

runtime/engine.cpp
```diff
@@ -342,13 +342,21 @@ void api::Engine::enable_module_mode(bool enable) {
   scriptLoader->enable_module_mode(enable);
 }
 
+JS::PersistentRootedObject SCRIPT_VALUE;
+JS::PersistentRootedObject api::Engine::script_value() {
```
tschneidereit, line 346:
```suggestion
HandleObject api::Engine::script_value() {
```

tschneidereit (approved):
Ooh, very exciting—and delightful how little change was needed!

I added two change suggestions, and am pretty surprised that things compiled without those (you're not supposed to pass `Rooted` things around, let alone `PersistentRooted` things). With those, we should very much land this!

guybedford:
You're right I just checked and the namespace properties are giving a rooting error actually, let me debug this tomorrow and update the PR with any changes to merge then.

guybedford:
I've now been able to get all the bindgen tests for ComponentizeJS to pass with these changes. I ended up extending the scope to a couple more items here as well.

This is ready for merge.

JakeChampion:
@tschneidereit is signed commits a requirement going forward for all contributions to StarlingMonkey?

guybedford:
I've gone ahead and signed all the commits here. The contents of the PR remain identical to last week.

runtime/script_loader.cpp
```diff
@@ -52,16 +67,16 @@ static char* resolve_path(const char* path, const char* base) {
 static bool load_script(JSContext *cx, const char *script_path, const char* resolved_path,
                         JS::SourceText<mozilla::Utf8Unit> &script);
 
-static JSObject* get_module(JSContext* cx, const char* path, const char* resolved_path,
+static JSObject* get_module(JSContext* cx, const char* specifier, const char* resolved_path,
                             const JS::CompileOptions &opts) {
-  RootedString path_str(cx, JS_NewStringCopyZ(cx, path));
-  if (!path_str) {
+  RootedString resolve_path_str(cx, JS_NewStringCopyZ(cx, resolved_path));
+  if (!resolve_path_str) {
```
tschneidereit, line 73:
```suggestion
  RootedString resolved_path_str(cx, JS_NewStringCopyZ(cx, resolved_path));
  if (!resolved_path_str) {
```

runtime/script_loader.cpp
```diff
@@ -52,16 +67,16 @@ static char* resolve_path(const char* path, const char* base) {
...
     return nullptr;
   }
 
   RootedValue module_val(cx);
-  RootedValue path_val(cx, StringValue(path_str));
-  if (!JS::MapGet(cx, moduleRegistry, path_val, &module_val)) {
+  RootedValue resolve_path_val(cx, StringValue(resolve_path_str));
+  if (!JS::MapGet(cx, moduleRegistry, resolve_path_val, &module_val)) {
```
tschneidereit, line 79:
```suggestion
  RootedValue resolved_path_val(cx, StringValue(resolved_path_str));
  if (!JS::MapGet(cx, moduleRegistry, resolved_path_val, &module_val)) {
```

runtime/script_loader.cpp
```diff
@@ -85,21 +100,13 @@ static JSObject* get_module(JSContext* cx, const char* path, const char* resolve
     return nullptr;
   }
 
-  if (!JS_DefineProperty(cx, info, "path", path_val, JSPROP_ENUMERATE)) {
+  if (!JS_DefineProperty(cx, info, "path", resolve_path_val, JSPROP_ENUMERATE)) {
```
tschneidereit, line 103:
```suggestion
  if (!JS_DefineProperty(cx, info, "path", resolved_path_val, JSPROP_ENUMERATE)) {
```

runtime/script_loader.cpp
```diff
@@ -85,21 +100,13 @@ static JSObject* get_module(JSContext* cx, const char* path, const char* resolve
...
-  // if (!addModuleToRegistry(cx, path, module)) {
-  //   return nullptr;
-  // }
-
-  if (!ModuleLink(cx, module)) {
-    return nullptr;
-  }
-
-  if (!MapSet(cx, moduleRegistry, path_val, module_val)) {
+  if (!MapSet(cx, moduleRegistry, resolve_path_val, module_val)) {
```
tschneidereit, line 109:
```suggestion
  if (!MapSet(cx, moduleRegistry, resolved_path_val, module_val)) {
```

runtime/script_loader.cpp
```diff
@@ -119,8 +126,28 @@ JSObject* module_resolve_hook(JSContext* cx, HandleValue referencingPrivate,
...
+  char* buf(js_pod_malloc<char>(parent_len));
+  auto result = JS_EncodeStringToUTF8BufferPartial(cx, parent_path, mozilla::Span<char>(buf, buf + parent_len));
+  if (result.isNothing()) {
+    js_free(buf);
+    return nullptr;
+  }
+  size_t written = std::get<1>(result.extract());
+  buf[written] = '\0';
+  char* resolved_path = resolve_path(path.get(), buf);
+  js_free(buf);
```
tschneidereit, line 150: It'd be nice if we could simplify this a bit, and I see a few different options for doing so (vaguely in descending order of neatness, I think, though I find myself waffling between the first two):

1. We could (in addition to storing it as a JS string under `path` as right now) store the path as a private value that holds the `char*` so it doesn't need to be re-encoded for all imports.
2. We could retain the base path in a simple linked list that gets added to every time we start parsing a file, and popped from when we're done with it, at which point we can free the chars.
3. We could clean up the code by using `core::encode` instead of implementing encoding here.

Separately from all that, we could pass the base path's length into `resolve_path` so it doesn't need to be re-derived there.

runtime/script_loader.cpp
```diff
@@ -237,7 +267,18 @@ bool ScriptLoader::load_top_level_script(const char *path, MutableHandleValue re
...
+    return JS_ExecuteScript(cx, script, result);
+  }
+  
+  if (!ModuleEvaluate(cx, module, result)) {
+    return false;
+  }
+  
+  // modules return the top-level await promise in the result value
+  // we don't currently support TLA, instead we reassign result
+  // with the module namespace
```
tschneidereit, line 280: Would you mind filing an issue for TLA? It really shouldn't be hard to support, so we should perhaps just do it before too long. (But of course not as a prerequisite to or part of this PR.)

tschneidereit (approved):
Thank you, this is fantastic!

It all looks great to me, except for some tiny tweaks I added as suggestions, and a slightly more substantive comment on some of the path resolution code. It'd be great to address that, but I can also do it as a commit pushed here, if you'd prefer.

runtime/script_loader.cpp
guybedford: I've added (3), and also added the explicit length pass.

tschneidereit (approved):
Thank you!

PR #12, opened 2024-03-19
wasi: latest wasi bindgen

guybedford:
This updates the current 0.2.0 WASI bindings to fix that UDP removed the `ipv6-only` methods.

Updates the WIT files to the latest WASI 0.2.0 WASI files, and the bindgen object file is fully regenerated. The wit-bindgen bindings are manually edited to avoid too much churn.


JakeChampion (approved):

tschneidereit (approved):
Thank you!

PR #23, opened 2024-04-12
add simple test runner and CI

guybedford:
Formalizes the existing smoke test into a CTest run, in a way that can hopefully scale to a few more tests, including a GitHub action workflow to run it.

First time I've used CTest so it can be improved, but it's a start.

Also fixes a build failure introduced in https://github.com/bytecodealliance/StarlingMonkey/pull/22.


tschneidereit (approved):
Thank you, this is great!

We'll want to have a more fully-fledged testing harness eventually, and we'll want to make our tests more self-contained (e.g. to not do calls to `example.com`), but for now this is much better than nothing!

PR #25, opened 2024-04-15
fix CI

guybedford:
This should fix the Wasmtime install, which I believe is due to the Rust toolchain lock to version `1.68.3`. In addition adds Wizer and wasm-tools installs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,13 +41,13 @@ jobs:
 
     - name: Install Wasmtime
       run: |
-        curl -LO https://github.com/bytecodealliance/wasmtime/releases/download/${tag}/wasmtime-${{ matrix.wasmtime }}-x86_64-linux.tar.xz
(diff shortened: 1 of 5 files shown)


tschneidereit:
These tools are all installed by the cmake script as well, so it'd be great if we could use those versions instead of building them from source for every run.

Perhaps a good way to enable this would be to introduce a CMake target that prints a `PATH` with the relevant directories added, and then have CI use that to modify the `PATH`?

guybedford:
Okay done, it seems like GitHub actions doesn't like altering the PATH for some reason, so I've used explicit WIZER and WASM_TOOLS env vars in the componentize script.

.github/workflows/main.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
+        wasmtime: [19.0.2]
+        wizer: [5.0.0]
+        wasm-tools: [1.203.0]
```
tschneidereit, line 27: Is this change still needed?

tschneidereit (approved):
This is great, thank you!

.github/workflows/main.yml
guybedford: Good catch, thanks.

Issue #14, opened 2024-03-20
Supporting top-level await


guybedford:
Top-level await support should be based on running the top-level application event loop to completion at preinitialization time to ensure the top-level has completed execution.

Errors can then be thrown as build errors.

Note that top-level `fetch` calls would effectively be build time fetches. Right now this would error but perhaps this may even be a useful feature if Wizering were to support it in future?

Note Fastly's JS Compute Runtime just landed support for TLA along the lines above.

guybedford:
Supported in #28.

PR #28, opened 2024-04-15
TLA

guybedford:
Adds initial support for TLA with some basic tests, resolving https://github.com/bytecodealliance/StarlingMonkey/issues/14.

There is still work to do on improving errors and unhandled rejections.

Also expands the test suite to better support error output testing to enable these follow-ups.

Note, this requires https://github.com/bytecodealliance/StarlingMonkey/pull/25 to land first for CI to pass.


README.md
```diff
@@ -47,22 +47,19 @@ Integration tests require [`Wasmtime`](https://wasmtime.dev/) to be installed (`
...
-cmake --build cmake-build-debug --parallel 8 --target test-cases
 CTEST_OUTPUT_ON_FAILURE=1 make -C cmake-build-debug test
 ```
 
 Individual tests can also be run from within the build folder using `ctest` directly:
 
 ```bash
-cd cmake-build-debug
-ctest smoke
+ctest --test-dir cmake-build-debug smoke
```
tschneidereit, line 56: These instructions _almost_ work, but not entirely:
1. `ctest` seems to only work after running `make -C cmake-build-debug test`. Is there a way to address that? If not, I think it'd be good to keep a build target for the tests that doesn't also run them
2. working around the above, I can get `ctest` to run, but it seems to always run all tests, instead of just the `smoke` test as you indicate here. Should this use `-R [pattern]` instead?

Separately, I think it'd be good to eventually support loading the tests at runtime instead of having to wizen every test individually. That's what the WPT harness does as well. We shouldn't use this for all tests, and ideally make it an option, but for most tests I think that'd be fine, and speed things up considerably.

runtime/engine.cpp
```diff
@@ -353,20 +353,39 @@ void api::Engine::abort(const char *reason) { ::abort(CONTEXT, reason); }
...
+  if (!scriptLoader->load_top_level_script(path, &ns, &tla_promise)) {
     return false;
   }
 
+  // Drive TLA as far as possible. Rejections during pre-initialization are
+  // treated as top-level exceptions. TLA may remain unresolved, in which case
+  // it will continue tasks at runtime. Rejections after pre-intialization remain
+  // unhandled rejections for now.
+  if (tla_promise.isObject()) {
+    if (!core::EventLoop::run_event_loop(this, 0, &tla_promise)) {
```
tschneidereit, line 366: We're not actually using the `result` parameter in `run_event_loop`, but if we were, this would overwrite the TLA promise.

Instead of taking this suggestion, please feel free to remove the `result` parameter :)
```suggestion
    RootedValue rval(cx);
    if (!core::EventLoop::run_event_loop(this, 0, &rval)) {
```

runtime/engine.cpp
```diff
@@ -353,20 +353,39 @@ void api::Engine::abort(const char *reason) { ::abort(CONTEXT, reason); }
...
-  if (!scriptLoader->load_top_level_script(path, &ns)) {
+  RootedValue tla_promise(cx);
+  if (!scriptLoader->load_top_level_script(path, &ns, &tla_promise)) {
     return false;
   }
 
+  // Drive TLA as far as possible. Rejections during pre-initialization are
+  // treated as top-level exceptions. TLA may remain unresolved, in which case
+  // it will continue tasks at runtime. Rejections after pre-intialization remain
+  // unhandled rejections for now.
```
tschneidereit, line 364: It'd be good to integrate this with the job queue draining below. Or rather, replace the latter with this wholesale, by unconditionally running `run_event_loop` and only then checking of `tla_promise` is an object, and then removing the `while (js::HasJobsPending(cx))` loop below.

tschneidereit (changes requested):

tschneidereit:
Oops, hit the button without leaving a comment. This looks really great, thank you!

I requested a few changes, but I think they should all be fairly minor.

README.md
guybedford: 1. `ctest` doesn't require `make -C cmake-build-debug test` to run first, just a normal `cmake --build cmake-build-debug`. I tested this again on a fresh build and it's definitely working.
2. Yes you're right about the `-R` pattern I've updated the example to note this.

Yes, runtime bulk tests are good, but also these tests can now catch all output and logs so are useful for fully isolated end to end test runs.

Perhaps the WPT tests and all the other unit tests can run as a single VM that just has everything for the fastest test run, and then we restrict these tests to only those that need to exercise the full end to end expectations.

runtime/engine.cpp
guybedford: Yeah I wasn't sure what the plan was here. Ok sure, I've replaced this with a pending exception check instead.

tschneidereit (approved):
Very nice, thank you!

runtime/js.cpp
```diff
@@ -42,16 +42,16 @@ bool initialize(const char *filename) {
 
   RootedValue result(engine.cx());
   bool success = engine.eval_toplevel(filename, &result);
-  success = success && engine.run_event_loop(&result);
+  success = success && engine.run_event_loop();
```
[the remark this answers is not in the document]
guybedford: Sure, nice to clean this up. Done.

PR #29, opened 2024-04-16
Support extension searching

guybedford:
Adds simple support for extension searching in imports.


tschneidereit (approved):
This looks good to me. I guess we'll at some point want to do something more fully-fledged, including supporting `.jsm` and module path searches, but all of that should be fully backwards compatible with this.

guybedford:
Yes, we should support the full Node.js algorithm here, and we can get there. Extension searching is strictly speaking not part of the modern Node.js resolution, but it's common enough in bundlers that it is worth enabling. It likely makes sense to disable it then in a breaking change eventually though, once enough of the JS ecosystem is no longer using extension searching.

PR #30, opened 2024-04-17
Support configuring enabled builtins using CMake options

tschneidereit:
This modifies the `add_builtin` CMake function to derive a CMake option from the name of the builtin, and only to only install the builtin if that option is set.

The README.md is also modified to document how to get a list of all available builtins, and whether they're enabled by default.

Note that this mechanism will automatically apply to CMake projects using StarlingMonkey, too.


guybedford (approved):
Tested this out and confirmed it is working for disabling features.

PR #32, opened 2024-04-19
Support builtin modules

guybedford:
This adds support for defining builtin modules from JS objects, which are then lazily turned into synthetic modules when imported, based on enumerating the own keys of the object and making those the exports of the synthetic module.

Useful for supporting custom internal APIs without needing to define globals.

### DIFF
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -38,6 +38,20 @@ public:
   HandleObject global();
 
   /**
+   * Define a new builtin import
+   * 
+   * The enumerable properties of the builtin object are used to construct
+   * a synthetic module namespace for the module.
+   * 
+   * The enumeration and getters are called only on the first import of
+   * the builtin, so that lazy getters can be used to lazily initialize
+   * builtins.
+   * 
+   * Once loaded, the instance is cached and reused as a singleton.
+   */
+  bool define_builtin_import(const char* id, HandleValue builtin);
+
+  /**
    * Treat the top-level script as a module or classic JS script.
    *
    * By default, the engine treats the top-level script as a module.
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -38,7 +38,7 @@ public:
   HandleObject global();
 
   /**
-   * Define a new builtin import
+   * Define a new builtin module
    * 
    * The enumerable properties of the builtin object are used to construct
    * a synthetic module namespace for the module.
@@ -49,7 +49,7 @@ public:
    * 
    * Once loaded, the instance is cached and reused as a singleton.
    */
-  bool define_builtin_import(const char* id, HandleValue builtin);
+  bool define_builtin_module(const char* id, HandleValue builtin);
 
   /**
    * Treat the top-level script as a module or classic JS script.
--- a/runtime/script_loader.cpp
+++ b/runtime/script_loader.cpp
@@ -4,6 +4,7 @@
 #include <cstdio>
 #include <iostream>
 #include <js/CompilationAndEvaluation.h>
+#include <jsfriendapi.h>
 #include <js/MapAndSet.h>
 #include <js/Value.h>
 #include <sys/stat.h>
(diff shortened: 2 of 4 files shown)


runtime/script_loader.cpp
```diff
@@ -166,15 +272,59 @@ JSObject* module_resolve_hook(JSContext* cx, HandleValue referencingPrivate,
   return get_module(cx, path.get(), resolved_path, opts);
 }
 
- ScriptLoader::ScriptLoader(JSContext *cx, JS::CompileOptions *opts) {
+bool module_metadata_hook(JSContext* cx, HandleValue referencingPrivate, HandleObject metaObject) {
```
guybedford, line 275: This hook sets the builtin as `import.meta.builtin` when the module is a builtin module.

runtime/script_loader.cpp
```diff
@@ -137,6 +139,99 @@ static JSObject* get_module(JSContext* cx, const char* specifier, const char* re
   return module;
 }
 
+static JSObject* get_builtin_module(JSContext* cx, HandleValue id, HandleObject builtin) {
```
guybedford, line 142: This does a synthetic module generation via source text. For example, for a JS builtin object like `{ env }`, the generated code looks like:

```js
const { e0: 'env' } = import.meta.builtin;
export { e0 as 'env' }
```

By always quoting the strings both in the destructuring and the reexport, we ensure all export names are supported.

Strictly speaking, this code doesn't yet support `'` in the export key currently, I can add that if we want to ensure 100% correctness.

include/extension-api.h
```diff
@@ -37,6 +37,20 @@ class Engine {
...
+   * The enumerable properties of the builtin object are used to construct
+   * a synthetic module namespace for the module.
+   * 
+   * The enumeration and getters are called only on the first import of
+   * the builtin, so that lazy getters can be used to lazily initialize
+   * builtins.
+   * 
+   * Once loaded, the instance is cached and reused as a singleton.
+   */
+  bool define_builtin_import(const char* id, HandleValue builtin);
```
guybedford, line 52: An example usage here might look like:

```c++
define_builtin_import("fastly:env", envObject);
```

runtime/script_loader.cpp
```diff
@@ -167,15 +272,59 @@ JSObject* module_resolve_hook(JSContext* cx, HandleValue referencingPrivate,
...
+  return true;
+}
+
+ScriptLoader::ScriptLoader(JSContext *cx, JS::CompileOptions *opts) {
   MOZ_ASSERT(!SCRIPT_LOADER);
   SCRIPT_LOADER = this;
   CONTEXT = cx;
   COMPILE_OPTS = opts;
   moduleRegistry.init(cx, JS::NewMapObject(cx));
+  builtinImports.init(cx, JS::NewMapObject(cx));
```
tschneidereit, line 301: I guess in principle we could keep to just having the ` moduleRegistry` map, and initializing the entries for uninitialized builtin modules to some magic value, which is then detected in the resolution hook and replaced with the initialized module.

That doesn't have to block landing this though, as I think we can live with the current approach perfectly well :)

tschneidereit (approved):
This is great, thank you so much!

include/extension-api.h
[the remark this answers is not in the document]
guybedford: Sure, renamed.

runtime/script_loader.cpp
guybedford: Yes, having a module registry with structured data I think would make sense as soon as we need more context in future. Possibly even an unordered map to structs with rooting.

guybedford: For now I'll leave this as-is though.

runtime/script_loader.cpp
```diff
@@ -137,6 +139,99 @@ static JSObject* get_module(JSContext* cx, const char* specifier, const char* re
...
+  GetPropertyKeys(cx, builtin, JSITER_OWNONLY, &props);
+
+  size_t length = props.length();
+  bool firstValue = true;
+  for (size_t i = 0; i < length; ++i) {
+    if (firstValue) {
+      firstValue = false;  
+    } else {
+      code += ", ";
+    }
```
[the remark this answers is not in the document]
guybedford: Yeah agreed, although I still prefer the proper formatting I must admit!

PR #33, opened 2024-04-19
Native stream layering

guybedford:
I'm currently building a custom request and response version against StarlingMonkey streams, which isn't currently working due to this hard dependence. Having the streams builtins include from request and response seems inverted in terms of the desired layering.

This removes that dependence, fixing the build layering for this use case.


tschneidereit (approved):
Yeah, that was messy—thank you for cleaning it up!

PR #34, opened 2024-04-19
Use resolved path in stacks

guybedford:
This fixes a bug where stacks contain the relative specifiers in errors like:

```
@./rel-path.js:17:32
app@index.js:69:25
@index.js:53:23
```

where instead it should be `dir/rel-path.js` in the stack.


tschneidereit (approved):
Makes sense!

PR #35, opened 2024-04-23
feat: integration tests, builtin tests, backtracking

guybedford:
This adds a new integration test runner variant, which takes a single multi-route component and then tests it against the expectations using the same expectation checking system.

Integration tests for the base64 builtins are then ported.

In addition, proper segment handling is added for path resolution in the loader, to now support backtracking.

Updates the CTests to run in parallel, supporting port assignment for running multiple servers at once. Also updates the CI to run tests in parallel.


runtime/script_loader.cpp
```diff
@@ -71,20 +71,56 @@ static const char* resolve_extension(const char* resolved_path) {
...
+
+  // copy the base in if used
+  size_t resolved_len = base_len;
+  if (path[0] == '/') {
+    resolved_len = 0;
+  } else {
+    strncpy(resolved_path, base, base_len);
+  }
+
+  // then iterate by segment of the path
```
tschneidereit, line 90: Can you expand this comment to explain what the loop actually does? It's not trivial to follow, and seems pretty relevant to getting the right behavior here.

tschneidereit (approved):
This LGTM, thanks! I asked for one change, but r=me with that addressed.

runtime/script_loader.cpp
guybedford: Sure, added some more comments.

Issue #37, opened 2024-04-24
setInterval stalls event loop entirely


guybedford:
Adding an arbitrary `setInterval(() => {})` inside of a server handler will stall the server handler from completing entirely.

We need to:

1. Ensure that other tasks continue running correctly
2. Ensure that the engine shutdown is properly managed when there are ongoing timer tasks

guybedford:
There is an integration test for this in the timers test suite added in https://github.com/bytecodealliance/StarlingMonkey/pull/38.

PR #38, opened 2024-04-24
Event loop reworking, timers tests

guybedford:
This PR reworks the event loop and adds integration tests for the timers builtins.

The `run_event_loop` function is extended to support `incr_event_loop_lifetime` and `decr_event_loop_lifetime` calls which determine how long the event loop lifetime should run for, and when this is zero only a single jobs processing is done.

This PR also updates the task class to support `poll` instead of `select` and makes `ready` a static method.

Resolves https://github.com/bytecodealliance/StarlingMonkey/issues/37, https://github.com/bytecodealliance/StarlingMonkey/issues/39.

### DIFF
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -40,17 +40,17 @@ public:
 
   /**
    * Define a new builtin module
-   * 
+   *
    * The enumerable properties of the builtin object are used to construct
    * a synthetic module namespace for the module.
-   * 
+   *
    * The enumeration and getters are called only on the first import of
    * the builtin, so that lazy getters can be used to lazily initialize
    * builtins.
-   * 
+   *
    * Once loaded, the instance is cached and reused as a singleton.
    */
-  bool define_builtin_module(const char* id, HandleValue builtin);
+  bool define_builtin_module(const char *id, HandleValue builtin);
 
   /**
    * Treat the top-level script as a module or classic JS script.
@@ -64,21 +64,33 @@ public:
   bool eval_toplevel(const char *path, MutableHandleValue result);
 
   /**
-   * Run the JS task queue and wait on pending tasks until there
-   * are no outstanding lifetimes to wait on.
+   * Run the async event loop as long as there's interest registered in keeping it running.
+   *
+   * Each turn of the event loop consists of three steps:
+   * 1. Run reactions to all promises that have been resolves/rejected.
+   * 2. Check if there's any interest registered in continuing to wait for async tasks, and
+   *    terminate the loop if not.
+   * 3. Wait for the next async tasks and execute their reactions
+   *
+   * Interest or loss of interest in keeping the event loop running can be signaled using the
+   * `Engine::incr_event_loop_interest` and `Engine::decr_event_loop_interest` methods.
+   *
+   * Every call to incr_event_loop_interest must be followed by an eventual call to
+   * decr_event_loop_interest, for the event loop to complete. Otherwise, if no async tasks remain
+   * pending while there's still interest in the event loop, an error will be reported.
    */
   bool run_event_loop();
 
   /**
-   * Add an event loop lifetime to track
+   * Add an event loop interest to track
    */
-  void incr_event_loop_lifetime();
+  void incr_event_loop_interest();
 
   /**
-   * Remove an event loop lifetime to track
+   * Remove an event loop interest to track
    * The last decrementer marks the event loop as complete to finish
    */
-  void decr_event_loop_lifetime();
+  void decr_event_loop_interest();
 
   /**
    * Get the JS value associated with the top-level script execution -
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -62,7 +62,8 @@ public:
   void enable_module_mode(bool enable);
   bool eval_toplevel(const char *path, MutableHandleValue result);
 
-  bool run_event_loop();
+  bool process_jobs();
+  bool process_async_tasks();
 
   /**
    * Get the JS value associated with the top-level script execution -
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -130,10 +130,9 @@ public:
   virtual void trace(JSTracer *trc) = 0;
 
   /**
-   * Poll for completion on the given async tasks
-   * A list of ready task indices is returned
+   * Select for the next available ready task, providing the oldest ready first.
    */
-  static std::vector<size_t> poll(std::vector<AsyncTask *> *handles);
+  static size_t select(std::vector<AsyncTask *> *handles);
 };
 
 } // namespace api
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -32,6 +32,12 @@ namespace api {
 class AsyncTask;
 
 class Engine {
+
+typedef int Lifetime;
+
+#define LIFETIME_NONE -1;
+#define LIFETIME_ALL -2;
+
 public:
   Engine();
   JSContext *cx();
@@ -62,8 +68,22 @@ public:
   void enable_module_mode(bool enable);
   bool eval_toplevel(const char *path, MutableHandleValue result);
 
-  bool process_jobs();
-  bool process_async_tasks();
+  /**
+   * Run the JS task queue and wait on pending tasks until there
+   * are no outstanding lifetimes to wait on.
+   */
+  bool run_event_loop();
+
+  /**
+   * Add an event loop lifetime to track
+   */
+  void incr_event_loop_lifetime();
+
+  /**
+   * Remove an event loop lifetime to track
+   * The last decrementer marks the event loop as complete to finish
+   */
+  void decr_event_loop_lifetime();
 
   /**
    * Get the JS value associated with the top-level script execution -
@@ -94,7 +114,6 @@ public:
 
   virtual bool run(Engine *engine) = 0;
   virtual bool cancel(Engine *engine) = 0;
-  virtual bool ready() = 0;
 
   [[nodiscard]] virtual PollableHandle id() {
     MOZ_ASSERT(handle_ != INVALID_POLLABLE_HANDLE);
@@ -103,13 +122,16 @@ public:
 
   virtual void trace(JSTracer *trc) = 0;
 
-  /// Returns the first ready `AsyncTask`.
-  ///
-  /// TODO: as an optimization, return a vector containing the ready head of the queue.
-  /// Note that that works iff the very first entry in the queue is ready, and then only
-  /// for the dense head of the queue, without gaps. This is because during processing
-  /// of the ready tasks, other tasks might become ready that should be processed first.
-  static size_t select(std::vector<AsyncTask *> *handles);
+  /**
+   * Poll for completion on the given async tasks
+   * A list of ready task indices is returned
+   */
+  static std::vector<size_t> poll(std::vector<AsyncTask *> *handles);
+
+  /**
+   * Returns whether or not the given task is ready
+   */
+  static bool ready(AsyncTask *task);
 };
 
 } // namespace api
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -134,11 +134,6 @@ public:
    * A list of ready task indices is returned
    */
   static std::vector<size_t> poll(std::vector<AsyncTask *> *handles);
-
-  /**
-   * Returns whether or not the given task is ready
-   */
-  static bool ready(AsyncTask *task);
 };
 
 } // namespace api
--- a/include/extension-api.h
+++ b/include/extension-api.h
@@ -33,11 +33,6 @@ class AsyncTask;
 
 class Engine {
 
-typedef int Lifetime;
-
-#define LIFETIME_NONE -1;
-#define LIFETIME_ALL -2;
-
 public:
   Engine();
   JSContext *cx();
--- a/runtime/event_loop.cpp
+++ b/runtime/event_loop.cpp
@@ -59,53 +59,62 @@ bool EventLoop::run_event_loop(api::Engine *engine, double total_compute) {
   queue.get().event_loop_running = true;
   JSContext *cx = engine->cx();
 
+  bool polled = false;
+  std::vector<size_t> ready_list;
+  size_t ready_idx;
+  size_t erasures_since_last_poll = 0;
+
   while (true) {
-    while (js::HasJobsPending(cx)) {
-      js::RunJobs(cx);
-      if (JS_IsExceptionPending(cx)) {
-        exit_event_loop();
-        return false;
-      }
-      if (interest_complete()) {
-        exit_event_loop();
-        return true;
-      }
-    }
+    // Run a microtask checkpoint
+    js::RunJobs(cx);
 
-    const auto tasks = &queue.get().tasks;
-
-    if (tasks->size() == 0) {
-      fprintf(stderr, "event loop error - both task and job queues are empty, but expected "
-                      "operations did not resolve");
+    if (JS_IsExceptionPending(cx)) {
       exit_event_loop();
       return false;
     }
-
-    auto ready = api::AsyncTask::poll(tasks);
-    const auto ready_len = ready.size();
-    // ensure ready list is monotonic
-    if (!std::is_sorted(ready.begin(), ready.end())) {
-      std::sort(ready.begin(), ready.end());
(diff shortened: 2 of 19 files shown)


guybedford:
I've extended this PR to refactor the event loop to support the desired liveness properties.

tschneidereit (changes requested):
Thank you for working on this! ❤️

I would like to suggest a different approach though, that I hope will be more composable and easier to get right for embedders/extension authors:

Instead of having to drive the event loop manually the way you implemented it here, what if the extension API would provide a way to signal ongoing interest in async event processing?

I'm imagining something like
```c++
Engine::increase_async_interest();
Engine::decrease_async_interest();
```

These would simply increment/decrement an internal counter, and the part of the event loop that has the commented-out check `if (!builtins::FetchEvent::is_active(fetch_event))` would be replaced by (in simplified pseudo-code) `if (interest_counter > 0)`.

This would then be used in the incoming request handler with something like this:
```c++
dispatch_fetch_event(fetch_event, &total_compute);
if (builtins::web::fetch::FetchEvent::is_active(fetch_event)) {
  ENGINE->increase_async_interest();
}
```

Additionally, `dec_count_handler` would be changed to include this:
```c++
if (!builtins::web::fetch::FetchEvent::is_active(fetch_event)) {
  ENGINE->decrease_async_interest();
}
```

Together, those should restore the `waitUntil` behavior, but in a way that's composable with other extensions that might keep the event loop alive for whatever other reasons.

If you feel like that's too much of a deviation from the path you chose and would prefer not to work on this, I'm happy to take it on next and send a new PR instead.

tschneidereit:
Correction on the above: the function to add the last code block to is [`dec_pending_promise_count`](https://github.com/bytecodealliance/StarlingMonkey/blob/9da56c2228b39055faac35e1e877e63f42c9d660/builtins/web/fetch/fetch_event.cpp#L37)

guybedford:
@tschneidereit good suggestion, I've gone ahead and made the change here. In addition I refactored the `select` host call to instead be a full poll, and also turned the `ready` hook into a static host call instead of a per-task method, resolving https://github.com/bytecodealliance/StarlingMonkey/issues/39.

Happy to take further feedback here as required.

include/extension-api.h
```diff
@@ -62,8 +63,23 @@ class Engine {
   void enable_module_mode(bool enable);
   bool eval_toplevel(const char *path, MutableHandleValue result);
 
+  /**
+   * Run the JS task queue and wait on pending tasks until there
+   * are no outstanding lifetimes to wait on.
+   */
   bool run_event_loop();
```
tschneidereit, line 70: I find `lifetime` pretty confusing as a term here. How would you feel about something like `interest`? With this comment modified like this, and other places (in code and comments) accordingly:
```suggestion
  /**
   * Run the async event loop as long as there's interest registered in keeping it running.
   *
   * Each turn of the event loop consists of three steps:
   * 1. Run reactions to all promises that have been resolves/rejected.
   * 2. Check if there's any interest registered in continuing to wait for async tasks, and terminate the loop if not.
   * 3. Wait for the next async task to become ready and execute it.
   *
   * Interest or loss of interest in keeping the event loop running can be signaled using the `Engine::raise_event_loop_interest` and `Engine::lower_event_loop_interest` methods.
   *
   * If no async tasks remain pending while there's still interest in the event loop, an error will be reported.
   */
  bool run_event_loop();
```

tschneidereit (changes requested):
This looks good overall, thank you!

Unfortunately there's a bunch of subtlety around the tasks processing, which I commented on in detail inline. Thank you for bearing with me on all this!

runtime/event_loop.cpp
```diff
@@ -38,38 +39,51 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
-    // Process async tasks.
-    if (has_pending_async_tasks()) {
-      auto tasks = &queue.get().tasks;
-      const auto index = api::AsyncTask::select(tasks);
+    // Unresolved lifetime error -
+    // if there are no async tasks, and the lifetime was not complete
+    // then we cannot complete the lifetime
+    if (tasks->size() == 0) {
+      return false;
+    }
```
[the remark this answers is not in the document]
guybedford: I've added errors to these paths.

runtime/event_loop.h
```diff
@@ -25,16 +25,18 @@ class EventLoop {
...
 
   /**
-   * Run the event loop until all async tasks have completed.
-   *
-   * Concretely, that means running a loop, whose body does two things:
-   * 1. Run all micro-tasks, i.e. pending Promise reactions
-   * 2. Run the next ready async task
-   *
-   * The loop terminates once both of these steps are null-ops.
+   * Run the event loop until all lifetimes are complete
```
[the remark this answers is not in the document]
guybedford: Added the event loop docs as suggested.

include/extension-api.h
guybedford: Renamed to interest, and added docs.

runtime/event_loop.cpp
```diff
@@ -38,38 +39,51 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
       auto task = tasks->at(index);
       if (!task->run(engine)) {
         return false;
       }
       tasks->erase(tasks->begin() + index);
+      if (lifetime_complete())
+        return true;
     }
-  } while (js::HasJobsPending(engine->cx()) || has_pending_async_tasks());
+  }
```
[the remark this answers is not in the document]
guybedford: Thanks for the point about the erasure. To address your points:

1. I've now added some code that checks that the ready list is sorted, and if it is not will first sort it, before running through the tasks. Task ordering should imply completion ordering so this seems fine to assume. Then with this monotonic assumption, we can handle the ready index by just decrementing the index by the number of erasures in the loop fixing this erasure offset issue.
2. I'm not sure we want to do a drain per task here for maximum parallelism. `task->run()` will do some immediate synchronous work, which in turn will have its own task queues and follow-up work. If we just drain directly all those sync task queues, we don't necessarily maximally interleave parallel work, which seems desireable in JS environments. Therefore I think the code as written is actually correct from a task management perspective, since the immediate reactions then properly interleave with other parallel work. For now, I've added a single `RunJobs` call along with this so that each individual task will get its immediate processing, followed by its immediate reactions, but my preference would still be to leave `RunJobs` out of here entirely. Well-written async code in JS should have stall points such that maximal interleaving is desired.
3. Another case to guard against is event loop reentrancy in this model. I've added a guard to ensure there is no reentrancy now. If you can think of any other edge cases please let me know.

tschneidereit (changes requested):
Getting much closer! :)

> I've now added some code that checks that the ready list is sorted, and if it is not will first sort it, before running through the tasks. Task ordering should imply completion ordering so this seems fine to assume. Then with this monotonic assumption, we can handle the ready index by just decrementing the index by the number of erasures in the loop fixing this erasure offset issue.

I'm unfortunately fairly certain that this still doesn't ensure the right order of events :( The scenario I outlined in my last review could still happen with this version: task `B` becoming ready while running task `A`, so it should be run before task `C`. I can't see a way to guarantee this without polling again.

And I do think that it's important to ensure the right order of event processing here, because everything else not only introduces the risk of violating implicit assumptions—it's also in violation of the [spec for event loop processing](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model): the spec wants us to always run the oldest ready task at the point where we choose the next task to run. In the `A, B, C` scenario, the current code wouldn't be guaranteed to do that.

> I'm not sure we want to do a drain per task here for maximum parallelism. task->run() will do some immediate synchronous work, which in turn will have its own task queues and follow-up work. If we just drain directly all those sync task queues, we don't necessarily maximally interleave parallel work, which seems desireable in JS environments.

The same spec cited above requires us to perform a microtask checkpoint right after every task, so I don't think we have all that much leeway here.

However, I also think that doing so will actually maximize parallelism and hence is desirable anyway.

here's a concrete scenario in which that'd be the case, and in which the order of operations would look very different, too. Say we have tasks `T1` and `T2`, with the following behavior:
1. Task `T1` resolves promise `P1` and takes 1ms to run.
2. Task `T2` resolves promise `P2` and takes 50ms to run.
3. The `then` handler `h1` for promise `P1` calls `fetch` to send HTTP request `H1` and takes 20ms to run.
6. The `then` handler  `h2` for promise `P2` also calls `fetch` to send HTTP request `H2` and takes 20ms to run.

If we don't drain the job queue, the order of execution will be
@0ms: `T1` -> @1ms: `T2` -> @51ms: `h1` -> `H1` -> @71ms: `h2` -> `H2`

If we _do_ drain the job queue, the order will instead be
@0ms: `T1` -> @1ms: `h1` -> `H1` -> @21ms: `T2` -> @71ms: `h2` -> `H2`

So in this scenario draining the job queue causes `H1` to be sent off 30ms earlier than not draining it, and should hence reduce end-to-end processing time for this application substantially.

> For now, I've added a single RunJobs call along with this so that each individual task will get its immediate processing, followed by its immediate reactions, but my preference would still be to leave RunJobs out of here entirely. Well-written async code in JS should have sufficient yielding to the event loop such that maximal interleaving is desired.

I hope the above explanation is convincing regarding the processing model. It'd probably be good to explicitly refer to the spec algorithm mentioned above, and document the code in terms of that algorithm's spec.

One thing I realized is that it'd probably be good to apply to refactorings to the code to make it easier to implement and follow:
1. extract out a `perform_microtask_checkpoint` method on `EventLoop` which encapsulates the iterative draining of the job queue and exception handling.
2. Move the logic of selecting the next task to run back into `AsyncTask::select`, have it return one task at a time, but only call the `poll` host call if we can't know which task will be ready next. I.e., in the code performing the `poll` host call, always return the first ready task, but if the oldest task is the one that's ready, also store the list of ready tasks following it (but only up to the first non-ready one.) Then only do the host call if that list is empty.

> Another case to guard against is event loop reentrancy in this model. I've added a guard to ensure there is no reentrancy now. If you can think of any other edge cases please let me know.

Great call!

This makes me realize that it'd probably be ideal if we use the event loop for all code execution, starting with top-level evaluation, which we'd express as a task. I don't think that's particularly important though, and certainly doesn't need to happen in this PR.

tschneidereit:
Oops, apologies for the accidental mentions to the folks with `@[digits]ms` handles!

runtime/event_loop.cpp
```diff
@@ -39,53 +40,74 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
-      if (JS_IsExceptionPending(cx))
+      if (JS_IsExceptionPending(cx)) {
+        exit_event_loop();
         return false;
-      if (lifetime_complete())
+      }
+      if (interest_complete()) {
+        exit_event_loop();
         return true;
+      }
```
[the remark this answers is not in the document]
guybedford: Good point, I've updated the implementation to ensure a succesfull completion on no interest even if there are no async tasks, so that no interest always corresponds to "one tick".

guybedford:
Thanks for going through the details and spec here, that makes sense. I suppose the main case I've been concerned about is the case where the host call might result in an indeterminate delay - but in the case of there being other tasks already, there doesn't seem to be any danger to running select or poll again, so this concern is not justified in this situation.

Given this, calling to the host select or poll gain seems absolutely fine, and I've gone ahead and implemented that here.

In terms of the separate optimization for contiguous indices, that optimization feels like an edge case though now in the sense that it's optimizing for a very specific situation just to remove one possible fast host call, whereas just reverting to the select host call seems like it reduces the memory footprint of the polling anyway. I implemented the AsyncTask::poll approach here first, but then in the last commit reverted back to AsyncTask::select, and without any contiguous optimization at all. If you'd prefer to keep the contiguous optimization I can revert the last commit back though.

For the microtask checkpoint, looking at the implementation of `runJobs` it clearly is a recursive task queue draining function - https://searchfox.org/mozilla-central/source/js/src/vm/JSContext.cpp#836. In addition searching for `HasJobsPending` is yielding no results for me - has this function been removed in latest SpiderMonkey or am I just not searching correctly here?

But the above makes me think that we **only** need to call `runJobs` to perform the spec-based microtask checkpoint, and don't need a loop or call to `HasJobsPending`, and thus don't need to abstract out a new `perform_microtask_checkpoint` either. I've gone ahead and implemented based on that assumption, but please double-check my logic on this.

runtime/event_loop.cpp
```diff
@@ -38,40 +40,63 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
+                        "operations did not resolve");
         return false;
       }
-      tasks->erase(tasks->begin() + index);
     }
-  } while (js::HasJobsPending(engine->cx()) || has_pending_async_tasks());
 
-  return true;
+    // Select the next task to run according to event-loop semantics of oldest-first.
+    size_t task_idx = api::AsyncTask::select(tasks);
```
lpereira, line 90: One of the things that can be done here so this doesn't block indefinitely if an I/O task never finishes, is to create a timer task if there are pending CPU-only things -- and make the event loop work only with `api::AsyncTask::select()`.

The timer task can be something like akin to a `setTimeout(runMicroTaskCheckpoints, 0)` if there are any microtasks to be executed.

This way it's easier to integrate pollable/selectable tasks with things that are purely CPU-bound, even if your `select()` primitive doesn't offer a timeout parameter.

runtime/event_loop.cpp
```diff
@@ -38,40 +40,63 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
+
+inline bool interest_complete() { return queue.get().interest_cnt == 0; }
 
+inline void exit_event_loop() { queue.get().event_loop_running = false; }
+
+bool EventLoop::run_event_loop(api::Engine *engine, double total_compute) {
+  if (queue.get().event_loop_running) {
+    fprintf(stderr, "cannot run event loop as it is already running");
+    return false;
+  }
```
lpereira, line 58: Reentrant event loops can be useful in some scenarios!  For instance, when implementing `alert()`, it's customary to show the dialog, enter the main loop again (so timers and whatnot are processed, but the JS that called the alert function is waiting for the event loop to finish), and leave the event loop when the dialog is dismissed.

Of course, for now, this is perfectly fine, but it's something to keep in mind if such functionality is ever needed.

runtime/event_loop.cpp
lpereira: (This would be akin to GLib's [`g_idle_add()`](https://docs.gtk.org/glib/func.idle_add.html).  In fact, it might be a good idea to read the documentation for the GLib main loop, as it's a tried and true design.)

tschneidereit: Unfortunately if I understand the proposal correctly that wouldn't be in compliance with how the DOM event loop works, which is what we're aiming for here. As mentioned in my last review, we always have to run the microtask checkpoint after every task, instead of enqueuing microtask handling for a later time.

runtime/event_loop.cpp
tschneidereit: I absolutely agree! And I actually think we'd be able to make the event loop reentrant without too much difficulty, too. I also think that it's ok to punt on that and address it in a future change :)

runtime/event_loop.cpp
```diff
@@ -38,40 +40,63 @@ bool EventLoop::cancel_async_task(api::Engine *engine, const int32_t id) {
...
-      auto tasks = &queue.get().tasks;
-      const auto index = api::AsyncTask::select(tasks);
-      auto task = tasks->at(index);
-      if (!task->run(engine)) {
+    const auto tasks = &queue.get().tasks;
+    size_t tasks_size = tasks->size();
+    if (tasks_size == 0) {
+      exit_event_loop();
+      // if there is no interest in the event loop at all, we always run one tick
+      if (interest_complete()) {
```
tschneidereit, line 80: Would this condition ever be met, given the check in line 70 above? If I'm not missing something and it indeed couldn't be met, maybe replace this with `MOZ_ASSERT(!interest_complete())`?

tschneidereit (approved):
This is looking great, thank you for your patience in working through it!

runtime/event_loop.cpp
guybedford: Good catch, thanks. Added the assertion instead.

PR #40, opened 2024-04-28
Remove underscore from base64DecodeTable

### LINKED ISSUE #36: atob inconsistency
For some reason `atob('__')` succeeds when it is supposed to fail.

In the base64 test suite added in https://github.com/bytecodealliance/StarlingMonkey/pull/35 this case was left out, but should be added back as soon as this can be fixed.

primoly:
Fixes #36


tschneidereit:
Thank you for this, @primoly! Can you say why this is the right fix? I *think* it is, based on the [table Firefox uses](https://searchfox.org/mozilla-central/source/xpcom/io/Base64.cpp#279), but would feel better if you gave a rationale for this change that we can then check.

guybedford (approved):

PR #43, opened 2024-05-07
Allow specifying dependencies from subprojects

### LINKED ISSUE #42: Using OpenSSL from a subproject
When trying to include OpenSSL from a subproject via, I'm getting an error for these includes, for example:

```
#include "openssl/evp.h"
```

gives that the include is not found.

Can we make it a feature of the subproject build that the OpenSSL headers are reachable? My CMake knowldge is fairly limited here - what would be involved to implement that?

lpereira:
This patch makes it possible for users of the `add_builtins()` CMake function to specify a per-builtin dependency on another target.  This has been used to fix #42.

### DIFF
--- a/cmake/add_builtin.cmake
+++ b/cmake/add_builtin.cmake
@@ -56,23 +56,28 @@ function(add_builtin)
         foreach(DEPENDENCY IN ITEMS ${DEPENDENCIES})
             get_target_property(TYPE ${DEPENDENCY} TYPE)
             if(NOT ${TYPE} MATCHES "UTILITY")
-                # A built-in can either depend on libraries implicitly
-                # linked by StarlingMonkey (i.e. depending on OpenSSL
-                # for your built-in when it only needs OpenSSL::Crypto),
-                # or on a library target.
+                # A built-in might specify a speficic dependency
+                # (e.g. "OpenSSL::Crypto").  Ensure that if this dependency
+                # is of the right type (i.e. libraries), this particular
+                # built-in is linked against this library.
+                # (CMake will take care of building e.g. "OpenSSL" first.)
                 target_link_libraries(${LIB_NAME} PRIVATE ${DEPENDENCY})
             endif()
 
-            # If a built-in requires a sub-dependency, e.g. the OpenSSL
-            # crypto library, we need to build the OpenSSL dependency
-            # first.
-            # The base name for the dependency is also how we're going
-            # to find the includes.
+            # If a dependency has the format "PackageName::SubDependency",
+            # e.g. "OpenSSL::Crypto", we need to grab only the PackageName
+            # to ensure that the target finds the libraries includes in the
+            # right place (in the next step).
             string(FIND "${DEPENDENCY}" "::" COLON_COLON)
             if(COLON_COLON GREATER_EQUAL 0)
                 string(SUBSTRING "${DEPENDENCY}" 0 COLON_COLON DEPENDENCY)
             endif()
 
+            # Only linking against a library isn't sufficient to build this
+            # built-in:  the compiler also needs to find the include files
+            # that this dependency brought.  CMake doesn't do this automatically,
+            # so give it a hand.
+            #
             # Not all dependencies will have their include files in this
             # location, but we can't, at this point, check if it exists,
             # because this code runs at configure time and this directory
--- a/cmake/add_builtin.cmake
+++ b/cmake/add_builtin.cmake
@@ -11,11 +11,11 @@ function(add_builtin)
         set(NS ${NS}::${NAME})
         set(DEFAULT_ENABLE ON)
     else()
-        cmake_parse_arguments(PARSE_ARGV 1 "" "DISABLED_BY_DEFAULT" "" "SRC;LINK_LIBS;INCLUDE_DIRS")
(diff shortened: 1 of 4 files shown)


guybedford:
I can confirm this resolves the issue for us.

cmake/add_builtin.cmake
```diff
@@ -50,6 +51,16 @@ function(add_builtin)
     target_link_libraries(${LIB_NAME} PRIVATE spidermonkey extension_api ${LINK_LIBS})
     target_link_libraries(builtins PRIVATE ${LIB_NAME})
     target_include_directories(${LIB_NAME} PRIVATE ${INCLUDE_DIRS})
+
+    if (DEPENDENCIES)
+	message(STATUS "${LIB_NAME} depends on ${DEPENDENCIES}")
```
tschneidereit, line 56:
```suggestion
	message(VERBOSE "${LIB_NAME} depends on ${DEPENDENCIES}")
```

tschneidereit (commented):
This looks good to me, thanks. One thing that's not entirely clear to me though is how it relates to the existing `LINK_LIBS` argument. That seems like it covers very similar ground? And if so, can we unify them somehow?

cmake/add_builtin.cmake
lpereira: Done.

lpereira:
Force-pushed changes that:

1. Remove `LINK_LIBS`
2. Establish `DEPENDENCIES` as a more general-purpose way of declaring the dependencies of built-ins
3. Installs OpenSSL in `${CMAKE_BINARY_DIR}/deps/OpenSSL`, so it's clearer that it's a dependency
4. Patches usages of `LINK_LIBS` in `builtins.cmake` to use `DEPENDENCIES` instead
5. Uses a `VERBOSE` message when printing information about a builtin's dependencies

guybedford:
Again I can confirm that the latest updates all work correctly in the Fastly branch.

guybedford (approved):

runtime/encode.h
```diff
@@ -2,7 +2,6 @@
 #define JS_COMPUTE_RUNTIME_ENCODE_H
 
 #include "host_api.h"
-#include "rust-url.h"
```
guybedford, line 5: I noticed that with this PR I needed to remove this unused include.

cmake/add_builtin.cmake
```diff
@@ -47,9 +47,46 @@ function(add_builtin)
...
+
+        add_dependencies(${LIB_NAME} ${DEPENDENCIES})
+
+        foreach(DEPENDENCY IN ITEMS ${DEPENDENCIES})
+            get_target_property(TYPE ${DEPENDENCY} TYPE)
+            if(NOT ${TYPE} MATCHES "UTILITY")
+                # A built-in can either depend on libraries implicitly
+                # linked by StarlingMonkey (i.e. depending on OpenSSL
+                # for your built-in when it only needs OpenSSL::Crypto),
+                # or on a library target.
```
tschneidereit, line 62: I think/hope that OpenSSL is actually not linked in if it's not depended on by any other target.

tschneidereit (commented):
I left two inline comments, which should be addressable with code comments, and then this should be ready to go!

(And apologies for the delay: I've been out sick for a while.)

cmake/add_builtin.cmake
lpereira: Done!

cmake/add_builtin.cmake
```diff
@@ -47,9 +47,46 @@ function(add_builtin)
...
+                string(SUBSTRING "${DEPENDENCY}" 0 COLON_COLON DEPENDENCY)
+            endif()
+
+            # Not all dependencies will have their include files in this
+            # location, but we can't, at this point, check if it exists,
+            # because this code runs at configure time and this directory
+            # will be created during compilation time.  This, however,
+            # doesn't cause compilation issues.
+            target_include_directories(${LIB_NAME}
+                       PRIVATE ${CMAKE_BINARY_DIR}/deps/${DEPENDENCY}/include)
```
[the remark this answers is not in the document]
lpereira: Done!

lpereira:
Rebased and force-pushed with the clarifications.

cmake/add_builtin.cmake
```diff
@@ -47,9 +47,51 @@ function(add_builtin)
...
+
+    if (DEPENDENCIES)
+        message(VERBOSE "${LIB_NAME} depends on ${DEPENDENCIES}")
+
+        add_dependencies(${LIB_NAME} ${DEPENDENCIES})
+
+        foreach(DEPENDENCY IN ITEMS ${DEPENDENCIES})
+            get_target_property(TYPE ${DEPENDENCY} TYPE)
+            if(NOT ${TYPE} MATCHES "UTILITY")
+                # A built-in might specify a speficic dependency
```
tschneidereit, line 59:
```suggestion
                # A built-in might specify a specific dependency
```

tschneidereit (approved):
Thank you, this looks great!

(For future reference, could you try to avoid force-pushing in PRs? Force-pushes make reviewing changes since the last review substantially harder, leading to longer review cycles. Thank you!)

PR #48, opened 2024-05-14
fix: crypto support & tests

guybedford:
This updates the crypto implementation with the PR in https://github.com/fastly/js-compute-runtime/pull/667 that was missed in the current implementation.

It also ports and verifies the full test suite is passing.


tschneidereit (approved):
Thank you, this is great!

I didn't do a thorough review, since this is mostly applying a pre-existing patch.

PR #52, opened 2024-05-28
Overhaul fetch implementation and host API

tschneidereit:
This PR contains a substantial overhaul of the `fetch` implementation, along with meaningful changes to how the host API works

On the latter, the most important change is that the representation of handles is now completely opaque to consumers: instead of using a `typedef` visible to users of the `host_api.h` header, the header just forward-declares a class, which hence can't be created or interacted with from outside the host API implementation.


tschneidereit:
@guybedford while this isn't completely done, and in dire need of cleanup, it is most of the way there, I think. I'm sharing it now because the host API changes are bound to impact you quite a bit. I hope they're for the better, and I at least found them to be necessary to get back to a point where I'm able to reason through the management and lifecycles of resource handles.

tschneidereit:
@guybedford I think this is now ready for review. The resource handling is much cleaned up, and I addressed all the regressions of WPT tests that make sense to do here, relative to the state of the JS Compute Runtime I started with.

(The only things missing that that passes are around some convenience constructors such as `Response.redirect()`, which I'm not tackling here.)

tschneidereit:
Thank you for letting me know, @nicoburniske! Fascinating, given that I ran a whole bunch of WPT tests around `fetch` that seemed to say it works. I'll investigate.

tschneidereit:
@nicoburniske thank you again for the report! Turns out I had read some of the WPT outputs wrong, and there indeed was a bunch more stuff to fix 😃

guybedford (approved):
I briefly tested this out in ComponentizeJS and can confirm fetch seems to behave correctly, down to some possible ComponentizeJS cases to work through.

guybedford:
One issue I did hit with this example:

```js
const res = await fetch('https://www.google.com');
const text = await res.text();
console.log(text);
```

is that it throws `(new TypeError("malformed UTF-8 character sequence at offset 1", ""))`

Is decompression properly supported?

tschneidereit:
@guybedford ah, interesting. The `fetch` implementation indeed doesn't do transparent decompression. But usually that shouldn't matter either, because we're not setting the `Accept-Encoding` header. Are you setting that, by any chance?

guybedford:
@tschneidereit because we virtualize on top of Node.js's fetch implementation it might be overriding that I expect, posted https://github.com/bytecodealliance/jco/issues/447 to investigate further.

guybedford:
Actually I spoke too soon. Running against a test server I'm getting mangled response body data. Headers are just `{"host":"localhost:8080","connection":"keep-alive"}`. But the body the test server is sending back doesn't seem to reflect at all. Again, this could be ComponentizeJS / Jco toolchain specific, but I'd have to dig into it more. We did pass all the WASI HTTP tests though, so I haven't seen any mangled response issues before, unless there are gaps in the test suite.

tschneidereit:
@guybedford I fixed a whole bunch of additional issues now. Would be great if you could take a look at the additional commits I pushed since your review. Then we should land this, and move everything else into additional PRs. (Of which I'll open a few in a moment, stacked on top of this one 😄)

runtime/engine.cpp
```diff
@@ -401,17 +448,35 @@ bool api::Engine::eval_toplevel(const char *path, MutableHandleValue result) {
...
+    JS::NonIncrementalGC(cx, JS::GCOptions::Normal, JS::GCReason::API);
+  }
 
   // Ignore the first GC, but then print all others, because ideally GCs
   // should be rare, and developers should know about them.
   // TODO: consider exposing a way to parameterize this, and/or specifying a
   // dedicated log target for telemetry messages like this.
   JS_SetGCCallback(cx, gc_callback, nullptr);
 
+  TOPLEVEL_EVALUATED = true;
```
guybedford, line 462: Strictly speaking in the case of TLA, the top-level hasn't been fully evaluated at this point. Perhaps we can call this something else like `execution_started`?

builtins/web/fetch/fetch_event.cpp
```diff
@@ -509,94 +528,58 @@ static void dispatch_fetch_event(HandleObject event, double *total_compute) {
   // LOG("Request handler took %fms\n", diff / 1000);
 }
 
-bool install(api::Engine *engine) {
-  ENGINE = engine;
-  FETCH_HANDLERS = new JS::PersistentRootedObjectVector(engine->cx());
+bool handle_incoming_request(host_api::HttpIncomingRequest * request) {
+  if (!ENGINE->toplevel_evaluated()) {
```
guybedford, line 532: What case is this branch for?

builtins/web/fetch/request-response.cpp
```diff
@@ -1740,8 +1726,8 @@ JSObject *Request::create(JSContext *cx, JS::HandleObject requestInstance, JS::H
   // Actually create the instance, now that we have all the parts required for
   // it. We have to delay this step to here because the wasi-http API requires
   // that all the request's properties are provided to the constructor.
-  auto request_handle = host_api::HttpOutgoingRequest::make(method, std::move(url), headers_handle);
-  RootedObject request(cx, create(cx, requestInstance, request_handle));
+  // auto request_handle = host_api::HttpOutgoingRequest::make(method, std::move(url), headers);
```
guybedford, line 1729: Is there a TODO here or can this be removed entirely?

guybedford (approved):
I'm still getting those weird issues with streaming when testing this in ComponentizeJS. I don't know if they are ComponentizeJS or Jco bugs though. I'm assuming streaming `res.text()` is properly captured in the WPT tests though? If so, what might be necessary to properly debug these cases is to try running the WPT using the Jco host instead of Wasmtime, to confirm the scenarios, before investigating if this is a ComponentizeJS issue, raising any upstream issue as necessary.

I have also tested this on Fastly's StarlingMonkey port and can confirm all still works correctly.

Would be great to land this now though and continue iterating with follow-ups as necessary.

runtime/engine.cpp
tschneidereit: I filed #64 to track improving this more generally

builtins/web/fetch/request-response.cpp
tschneidereit: Good catch, thank you! This, as well as the comment above, can be removed entirely.

builtins/web/fetch/fetch_event.cpp
tschneidereit: I extracted this out into a function with a short doc comment explaining the purpose.

tschneidereit:
Thank you for the review! I pushed a commit addressing your comments. While there's a bunch of code motion in that commit, it's all trivial, so I'll go ahead and land this.

Issue #50, opened 2024-05-24
Release build is broken


guybedford:
The release build is currently not working correctly. This may have happened in the latest SpiderMonkey update.

Build error reported:

```
[ 90%] Linking CXX executable starling.wasm
wasm-ld: error: duplicate symbol: encoding_for_label
>>> defined in librust_encoding.a(encoding_c-6718b175c3733dba.encoding_c.5cbbfd59fed8dac6-cgu.0.rcgu.o)
>>> defined in /Users/gbedford/Projects/StarlingMonkey/deps/cpm_cache/spidermonkey-release/0724b80a62196812717c8fbd9286c717d83c4113/spidermonkey-release/lib/libjsrust.a(jsrust-53a982b7a59e3f77.jsrust.86da07e83e4d91bd-cgu.0.rcgu.o)

wasm-ld: error: duplicate symbol: encoding_for_label_no_replacement
>>> defined in librust_encoding.a(encoding_c-6718b175c3733dba.encoding_c.5cbbfd59fed8dac6-cgu.0.rcgu.o)
>>> defined in /Users/gbedford/Projects/StarlingMonkey/deps/cpm_cache/spidermonkey-release/0724b80a62196812717c8fbd9286c717d83c4113/spidermonkey-release/lib/libjsrust.a(jsrust-53a982b7a59e3f77.jsrust.86da07e83e4d91bd-cgu.0.rcgu.o)

wasm-ld: error: duplicate symbol: encoding_for_bom
>>> defined in librust_encoding.a(encoding_c-6718b175c3733dba.encoding_c.5cbbfd59fed8dac6-cgu.0.rcgu.o)
>>> defined in /Users/gbedford/Projects/StarlingMonkey/deps/cpm_cache/spidermonkey-release/0724b80a62196812717c8fbd9286c717d83c4113/spidermonkey-release/lib/libjsrust.a(jsrust-53a982b7a59e3f77.jsrust.86da07e83e4d91bd-cgu.0.rcgu.o)

wasm-ld: error: duplicate symbol: encoding_name
>>> defined in librust_encoding.a(encoding_c-6718b175c3733dba.encoding_c.5cbbfd59fed8dac6-cgu.0.rcgu.o)
[...]

PR #53, opened 2024-05-28
fix: patch for release build

guybedford:
After two days of running every build configuration change I can think of I just can't figure out the root cause of https://github.com/bytecodealliance/StarlingMonkey/issues/50.

I cloned spidermonkey-wasi-embedding and ran a release build against the same Rust toolchain used in js-compute-runtime and was able to get js-compute-runtime to build correctly against it. But StarlingMonkey just won't link against its libjsrust.a in release mode.

This implements a workaround that at least gets the release build to work, by copying the debug build's libjsrust.a instead. Perhaps it points to something in the StarlingMonkey configuration that isn't properly updating the Rust build to be a release instead of a debug build in the right way, I'm just not sure how duplicate function errors get determined in the linker and if it has anything to do with structural code similarity or not.

I'm going to stop running builds all day, and just use this patch. Help figuring out the root case on https://github.com/bytecodealliance/StarlingMonkey/issues/50 would be very welcome.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,16 +22,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
+        mode: [debug, release]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-
-    - name: Install Rust 1.68.2
-      run: |
-        rustup toolchain install 1.68.2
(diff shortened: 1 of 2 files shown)


guybedford:
This PR also adds the release build into CI to avoid future regressions.

guybedford:
While this resolves https://github.com/bytecodealliance/StarlingMonkey/issues/50, it's only a temporary workaround until the root cause in the Cmake corrosion configuration can be found.

PR #54, opened 2024-05-28
fix: release build on fetch rework

guybedford:
I needed these changes to get the release build to work on https://github.com/bytecodealliance/StarlingMonkey/pull/52, but can confirm this works on js-compute-runtime and ComponentizeJS.


tschneidereit (approved):
Thank you!

PR #58, opened 2024-06-05
feat: support task deadlines

guybedford:
Fastly's js-compute-runtime does not support timer subscriptions, and instead virtual timers are used which check their deadlines as part of the normal event loop operation.

This PR implements a concept of deadlines for all tasks, so that the host select implementation can use those deadlines to progress task deadlines for hosts without timing subscriptions.

Feedback very welcome. This is one of the last pieces needed for the migration to StarlingMonkey for js-compute-runtime, with the migration passing 99% of all other tests down to this feature.

### DIFF
--- a/builtins/web/timers.cpp
+++ b/builtins/web/timers.cpp
@@ -68,6 +68,10 @@ public:
     return true;
   }
 
+  [[nodiscard]] uint64_t deadline(api::Engine *engine) override {
+    return deadline_;
+  }
+
   void trace(JSTracer *trc) override {
     TraceEdge(trc, &callback_, "Timer callback");
     for (auto &arg : arguments_) {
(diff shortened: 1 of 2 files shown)


tschneidereit (commented):
I'm probably missing some key aspects here, that make what I'm about to say invalid. With that said:

It seems to me like this could all be implemented outside of the core event loop mechanism, in two parts:
1. Instead of having the new additional loop over the tasks inside the main event loop, have the js-compute implementation of `api::AsyncTask::select` include that loop.
2. Add a `deadline` field to the `AsyncTask` subclasses, which gets set by the js-compute `api::AsyncTask::select` implementation, which then returns that task.

This way, the main event loop and the `AsyncTask` interface could stay unchanged. But again, I'm probably missing something that makes this unviable.

guybedford:
I've updated this PR to just implement the `deadline()` virtual function and provide it on the timer task as you've suggested.

tschneidereit (approved):
Excellent!

PR #60, opened 2024-06-05
refactor: select to take a tasks vector reference

guybedford:
Updates the `HostApi::select` interface to take a reference to a vector instead of a pointer.

### DIFF
--- a/runtime/event_loop.cpp
+++ b/runtime/event_loop.cpp
@@ -84,7 +84,7 @@ bool EventLoop::run_event_loop(api::Engine *engine, double total_compute) {
     }
 
     // Select the next task to run according to event-loop semantics of oldest-first.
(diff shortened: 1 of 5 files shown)


runtime/event_loop.cpp
```diff
@@ -84,7 +84,7 @@ bool EventLoop::run_event_loop(api::Engine *engine, double total_compute) {
     }
 
     // Select the next task to run according to event-loop semantics of oldest-first.
-    size_t task_idx = api::AsyncTask::select(tasks);
+    size_t task_idx = api::AsyncTask::select(*tasks);
```
elliottt, line 87: I really like that the assumption about `tasks` being non-null is pushed out to here, rather than being something that `select` might need to care about :+1:

elliottt (approved):
:tada:

tschneidereit (approved):

PR #61, opened 2024-06-18
Improve handling of GC managed array buffers

tschneidereit:
Specifically, some of the `noGC` guards had issues that could lead to panics during exception throwing, and some of the loops over multiple buffers where overly complex.


tschneidereit:
I'll change this to target `main` before landing, and will wait for the `fetch-rework` branch to land before doing so.

guybedford (approved):

PR #63, opened 2024-06-19
timers: allow clearTimeout and clearInterval during initialization

guybedford:
Strictly speaking, `clearTimeout` and `clearInterval` are valid noops when passed invalid values, therefore there is no reason to ban them at initialization time.

This behaviour is for some reason needed for Fastly's web platform test suite as well.

It does no harm, so I see no reason not to land this provided that `setTimeout` and `setInterval` remain protected.


tschneidereit (approved):
Makes sense to me, yeah. We should eventually also support other timers during initialization, but that's more involved because we'd need to use a WASIp1 based polling mechanism.

PR #66, opened 2024-06-21
Bugfixes for `setTimeout` and `setInterval`

tschneidereit:
Before this patch, timer (i.e., `setTimeout`/`setInterval`) returned the pollable handles as IDs. This is problematic, because pollable handles are reused, leading to content potentially clearing the wrong timer.

Additionally, the code wasn't robust against a timer being cleared in its own callback, leading to a failing assertion.

I also added a few tweaks to the test code that I did while writing the test for this.

### DIFF
--- a/tests/integration/test-server.js
+++ b/tests/integration/test-server.js
@@ -35,6 +35,21 @@ export function serveTest (handler) {
             curSubtest = null;
             subtestCnt++;
           }
+        },
+        async asyncTest (name, subtest) {
+          if (!filter(name))
+            return;
+          curSubtest = name;
+          let resolve, reject;
+          let promise = new Promise((res, rej) => {
+            resolve = res;
+            reject = rej;
+          });
+          subtest(resolve, reject);
+          evt.waitUntil(promise);
+          await promise;
+          curSubtest = null;
+          subtestCnt++;
         }
       });
     }
--- a/builtins/web/timers.cpp
+++ b/builtins/web/timers.cpp
@@ -6,6 +6,7 @@
 #include <host_api.h>
 #include <iostream>
 #include <list>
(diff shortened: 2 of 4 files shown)


tests/integration/test-server.js
```diff
@@ -35,6 +35,21 @@ export function serveTest (handler) {
...
+          if (!filter(name))
+            return;
+          curSubtest = name;
+          let resolve, reject;
+          let promise = new Promise((res, rej) => {
+            resolve = res;
+            reject = rej;
+          });
+          subtest(resolve, reject);
+          evt.waitUntil(promise);
```
guybedford, line 49: It would be interesting to know why this is needed since strictly speaking the test async function itself is blocking the main event handler (due to the `await t.asyncTest(...)`), so this shouldn't actually be needed?

guybedford (approved):
Thanks, I was just hitting WPT failures for timers, so this is much needed!

tschneidereit:
> Thanks, I was just hitting WPT failures for timers, so this is much needed!

I was working through probably those same failures, yeah :)

tests/integration/test-server.js
tschneidereit: It's not strictly speaking necessary, you're right. I at first misunderstood how async tests work, and had a hard time debugging why things didn't work, which is why I introduced this explicit `asyncTest` method.

I think it'd actually be nice to remove the async support from the normal `test`, to make it harder to misuse that one—but that'd of course require converting all existing tests.

guybedford: The API pattern we follow is exactly the same as Node.js's native test runner - https://nodejs.org/docs/latest/api/test.html#test-runner.

PR #69, opened 2024-06-21
feat: allow direct base64 call against HandleValue

guybedford:
Adds back the ability to call `convertJSValueToByteString` for a `HandleValue` by declaring the header for the original function variant. Necessary for Fastly's `statusText` setter to pass the web platform tests, because converting to a string first was altering unsupported characters.

### DIFF
--- a/builtins/web/base64.cpp
+++ b/builtins/web/base64.cpp
@@ -5,7 +5,7 @@ namespace builtins {
 namespace web {
 namespace base64 {
 
-JS::Result<std::string> convertJSValueToByteString(JSContext *cx, JS::Handle<JS::Value> v) {
+JS::Result<std::string> valueToJSByteString(JSContext *cx, JS::Handle<JS::Value> v) {
   JS::RootedString s(cx);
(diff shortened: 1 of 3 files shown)


tschneidereit (approved):
I hadn't really looked at this code before. Obviously adding this declaration makes sense, and I'm okay with landing this as-is.

Ideally though, it'd be great to rename at least the other export, or perhaps both. I think something like `ToJSByteString` would work well for both.

Also, it took me a bit to understand what this does: I first thought it'd base64 encode the input, where really what it does is coerce the input value to a latin1 string iff the result of stringifying it allows for that to happen losslessly. Having a comment explaining that would be great.

However, none of that needs to block landing this if you'd prefer getting on with things.

guybedford:
I've renamed these to `valueToJSByteString` and `stringToJSByteString` respectively, let me know if that works for you.

tschneidereit:
Sure, yeah. I think we could give them the same name and rely on overloading resolution, but this works, too.

PR #70, opened 2024-06-21
deps: use latest rust url

guybedford:
This updates to the latest `url` crate dependency, fixing a couple of web platform tests in the process (which were required to get Fastly's WPT to pass, along with https://github.com/bytecodealliance/StarlingMonkey/pull/69).


tschneidereit (approved):
Thanks!

PR #71, opened 2024-06-22
fix: support non-unique handles for timers

guybedford:
This was necessary to get one of the WPT cases to pass for Fastly's implementation.

Our implementation of checking task deadlines for timer tasks uses a redundant singular invalid handle (`-2`) to indicate a timer handle for a non-existent subscription. If we gave every timer a unique handle we would be using up the real handle ID space.

But the `cancel_task` currently assumes handle uniqueness when finding the handle.

This supports non-unique handles for tasks by taking the task itself to `cancel_task` instead of its handle to look it up from, storing timers as a map from timer ids to task pointers.

I can confirm this is working correctly in the WPT test case for Fastly.

### DIFF
--- a/builtins/web/timers.cpp
+++ b/builtins/web/timers.cpp
@@ -111,7 +111,7 @@ public:
       return false;
     }
 
-    ENGINE->remove_async_task(timer_ids_[timer_id]);
+    ENGINE->cancel_async_task(timer_ids_[timer_id]);
     timer_ids_.erase(timer_id);
     return true;
   }
--- a/builtins/web/timers.cpp
+++ b/builtins/web/timers.cpp
@@ -12,13 +12,30 @@
 #define S_TO_NS(s) ((s) * 1000000000)
 #define NS_TO_MS(ns) ((ns) / 1000000)
 
+namespace {
+
+class TimersMap {
+public:
+  std::map<int32_t, api::AsyncTask*> timer_ids_ = {};
+  int32_t next_timer_id = 1;
+  void trace(JSTracer *trc) {
+    for (std::map<int32_t, api::AsyncTask*>::iterator i = timer_ids_.begin(); i != timer_ids_.end(); i++) {
+      api::AsyncTask *timer = i->second;
+      timer->trace(trc);
+    }
+  }
+};
+
+}
+
+static TimersMap *TIMERS_MAP;
 static api::Engine *ENGINE;
 
 class TimerTask final : public api::AsyncTask {
   using TimerArgumentsVector = std::vector<JS::Heap<JS::Value>>;
 
-  static std::map<int32_t, api::AsyncTask*> timer_ids_;
-  static int32_t next_timer_id;
+  // static std::map<int32_t, api::AsyncTask*> timer_ids_;
+  // static int32_t next_timer_id;
 
   int32_t timer_id_;
   int64_t delay_;
@@ -40,8 +57,8 @@ public:
     }
 
     handle_ = host_api::MonotonicClock::subscribe(deadline_, true);
-    timer_id_ = next_timer_id++;
-    timer_ids_.emplace(timer_id_, this);
+    timer_id_ = TIMERS_MAP->next_timer_id++;
+    TIMERS_MAP->timer_ids_.emplace(timer_id_, this);
   }
 
   [[nodiscard]] bool run(api::Engine *engine) override {
@@ -68,13 +85,13 @@ public:
       host_api::MonotonicClock::unsubscribe(handle_);
     }
 
-    if (timer_ids_.contains(timer_id_)) {
+    if (TIMERS_MAP->timer_ids_.contains(timer_id_)) {
       if (repeat_) {
         deadline_ = host_api::MonotonicClock::now() + delay_;
         handle_ = host_api::MonotonicClock::subscribe(deadline_, true);
         engine->queue_async_task(this);
       } else {
-        timer_ids_.erase(timer_id_);
+        TIMERS_MAP->timer_ids_.erase(timer_id_);
       }
     }
 
@@ -82,7 +99,7 @@ public:
   }
 
   [[nodiscard]] bool cancel(api::Engine *engine) override {
-    if (!timer_ids_.contains(timer_id_)) {
+    if (!TIMERS_MAP->timer_ids_.contains(timer_id_)) {
       return false;
     }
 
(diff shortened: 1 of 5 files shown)


builtins/web/timers.cpp
```diff
@@ -113,7 +117,7 @@ class TimerTask final : public api::AsyncTask {
   }
 };
 
-std::map<int32_t, PollableHandle> TimerTask::timer_ids_ = {};
```
tschneidereit, line 116: As-is, this isn't GC-safe: if a moving GC were to happen and nothing else were to update the references inside `TimerTask`, they would become invalid.

In practice this situation can probably not currently arise, because any task included here should also be in the event queue. But that seems dangerous to rely on in the face of potential refactorings later on.

One way I can see to make this work is to introduce a new class, `TimersMap` which has a `trace` method and holds the `timer_ids_` field. We can then add a singleton instance of that as a `static PersistentRooted<TimersMap>` member of `TimerTask` and have the `trace` method iterate over the map to trace through the contained tasks.

tschneidereit (changes requested):
If I understand the problem correctly, all timers in Fastly's implementation have `PollableHandle == -2` because of how timeouts are translated, right?

If so, that makes sense to me, yeah.

The tricky bit, and why I didn't immediately implement things this way, is that the `TimerTask`s in the map need to be properly rooted—see the comment in-line.

guybedford:
> If I understand the problem correctly, all timers in Fastly's implementation have PollableHandle == -2 because of how timeouts are translated, right?

Yes exactly.

I've implemented a very naive singleton and trace model here as a start. Further feedback very much welcome.

builtins/web/timers.cpp
```diff
@@ -12,13 +12,30 @@
 #define S_TO_NS(s) ((s) * 1000000000)
 #define NS_TO_MS(ns) ((ns) / 1000000)
 
+namespace {
+
+class TimersMap {
+public:
+  std::map<int32_t, api::AsyncTask*> timer_ids_ = {};
```
tschneidereit, line 19: Given the new meaning, I'd prefer changing the name of this map, too:

```suggestion
  std::map<int32_t, api::AsyncTask*> timers_ = {};
```

builtins/web/timers.cpp
```diff
@@ -12,13 +12,30 @@
...
+static TimersMap *TIMERS_MAP;
 static api::Engine *ENGINE;
 
 class TimerTask final : public api::AsyncTask {
   using TimerArgumentsVector = std::vector<JS::Heap<JS::Value>>;
 
-  static std::map<int32_t, PollableHandle> timer_ids_;
-  static int32_t next_timer_id;
+  // static std::map<int32_t, api::AsyncTask*> timer_ids_;
+  // static int32_t next_timer_id;
```
tschneidereit, line 39:
```suggestion
```

builtins/web/timers.cpp
```diff
@@ -12,13 +12,30 @@
...
 
+namespace {
+
+class TimersMap {
+public:
+  std::map<int32_t, api::AsyncTask*> timer_ids_ = {};
+  int32_t next_timer_id = 1;
+  void trace(JSTracer *trc) {
+    for (std::map<int32_t, api::AsyncTask*>::iterator i = timer_ids_.begin(); i != timer_ids_.end(); i++) {
+      api::AsyncTask *timer = i->second;
```
tschneidereit, line 23:
```suggestion
    for (auto& [id, timer] : timers_) {
```

builtins/web/timers.cpp
```diff
@@ -12,13 +12,30 @@
...
+      timer->trace(trc);
+    }
+  }
+};
+
+}
+
+static TimersMap *TIMERS_MAP;
```
tschneidereit, line 31: Wrapping this in a `PersistentRooted` makes it be traced automatically:

```suggestion
static PersistentRooted<TimersMap*> TIMERS_MAP;
```

builtins/web/timers.cpp
```diff
@@ -100,22 +121,20 @@ class TimerTask final : public api::AsyncTask {
     for (auto &arg : arguments_) {
       TraceEdge(trc, &arg, "Timer callback arguments");
     }
+    TIMERS_MAP->trace(trc);
```
tschneidereit, line 124: While this works, it means that we call `TIMERS_MAP->trace` from every timer, while that method itself traces through the timers. Also, with the `PersistentRooted` thing above, it's not needed anymore.
```suggestion
```

builtins/web/timers.cpp
```diff
@@ -194,6 +213,7 @@ constexpr JSFunctionSpec methods[] = {
 
 bool install(api::Engine *engine) {
   ENGINE = engine;
+  TIMERS_MAP = new TimersMap();
```
tschneidereit, line 216:
```suggestion
  TIMERS_MAP.init(engine->cx(), new TimersMap());
```

tschneidereit (approved):
Looks good! I left a bunch of suggestions which should be applied (and then a little bit of other tweaks as follow-ups to those, because I didn't add suggestions for all the places where `s/timer_ids_/timers_/g` needs to happen), but with those in, this'll be ready to go!

builtins/web/timers.cpp
guybedford: This doesn't seem to build, giving the error:

```
/Users/gbedford/Projects/StarlingMonkey/deps/cpm_cache/spidermonkey-debug/83add297093a844dd3ba963092e46338080a42b3/spidermonkey-debug/include/js/RootingAPI.h:1399:41: note: in instantiation of template type alias 'RootedTraits' requested here
class PersistentRooted : public detail::RootedTraits<T>::PersistentBase,
                                        ^
/Users/gbedford/Projects/StarlingMonkey/builtins/web/timers.cpp:30:37: note: in instantiation of template class 'JS::PersistentRooted<(anonymous namespace)::TimersMap *>' requested here
static PersistentRooted<TimersMap*> TIMERS_MAP;
```

builtins/web/timers.cpp
```diff
@@ -194,6 +208,7 @@ constexpr JSFunctionSpec methods[] = {
 
 bool install(api::Engine *engine) {
   ENGINE = engine;
+  TIMERS_MAP.init(engine->cx(), new TimersMap());
```
tschneidereit, line 211:
```suggestion
  TIMERS_MAP.init(engine->cx(), js::MakeUnique<TimersMap>());
```

builtins/web/timers.cpp
```diff
@@ -12,14 +12,27 @@
...
+  void trace(JSTracer *trc) {
+    for (auto& [id, timer] : timers_) {
+      timer->trace(trc);
+    }
+  }
+};
+
+}
+
+static PersistentRooted<TimersMap*> TIMERS_MAP;
```
tschneidereit, line 30:
```suggestion
static PersistentRooted<js::UniquePtr<TimersMap>> TIMERS_MAP;
```

tschneidereit (approved):
Oh huh, seems like that doesn't work with raw pointers, sorry!

The suggested edits do make it work, because tracing is implemented for `js::UniquePtr`.

PR #72, opened 2024-06-22
fix: zero coercion timeout removal bug

guybedford:
This fixes a bug where `clearTimeout(null)` could clear a valid timeout if one existed at timer id `0`, by starting the timer ids from `1`. Affected a real web platform test!


tschneidereit (approved):
I should've thought of that, apologies

PR #73, opened 2024-06-23
Change all error reporting to use `api::throw_error`

tschneidereit:
I went through the existing errors with a pretty fine comb and cleaned things up where it makes sense, instead of just moving things over mechanically. As a result, a few more WPT tests pass, and in general things are a bunch cleaner now, I think.


builtins/web/structured-clone.cpp
```diff
@@ -45,6 +45,8 @@ JSObject *ReadStructuredClone(JSContext *cx, JSStructuredCloneReader *r,
   return params_obj;
 }
 
+DEF_ERR(CloningFailed, JSEXN_TYPEERR, "The object could not be cloned", 0)
```
guybedford, line 48: I believe this is supposed to be a `DataCloneError` DOM exception per https://html.spec.whatwg.org/multipage/structured-data.html#structured-cloning.

builtins/web/timers.cpp
```diff
@@ -131,9 +131,8 @@ template <bool repeat> bool setTimeout_or_interval(JSContext *cx, const unsigned
   }
 
   if (!(args[0].isObject() && JS::IsCallable(&args[0].toObject()))) {
-    JS_ReportErrorASCII(cx, "First argument to %s must be a function",
-                        repeat ? "setInterval" : "setTimeout");
-    return false;
+    return api::throw_error(cx, api::Errors::WrongType, repeat ? "setInterval" : "setTimeout",
+      "first argument", "be a function");
```
guybedford, line 135: Strictly speaking, I'm not sure this is a required error for timers per https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers.

guybedford (approved):

builtins/web/crypto/crypto-errors.h
```diff
@@ -0,0 +1,12 @@
+#ifndef CRYPTO_ERRORS_H
+#define CRYPTO_ERRORS_H
+
+namespace CryptoErrors {
+DEF_ERR(InvalidJwk, JSEXN_TYPEERR, "crypto.subtle.importkey: The JWK member '{0}' was not '{1}'", 2)
```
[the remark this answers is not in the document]
tschneidereit: good catch, replaced all usages

builtins/web/timers.cpp
tschneidereit: the reason for this exception is that we don't currently support the invocation mode where a source string is passed in instead of a function. There's no fundamental reason not to do so, but it's not something I want to tackle here.

tschneidereit:
@guybedford I changed a number of test expectations to match the new errors. I think those are all at least as correct as before, and in some cases clearly more so, so I would just land this.

Another thing I did however is a bit more contentious perhaps: I removed all the error messages from the test expectations. Error messages are a notoriously fickle thing to test against, and I think we're better served with checking the error type and otherwise leaning on breadth of coverage to catch all the different reasons for why an exception might be thrown.

If you strongly disagree, we should discuss this more, though!

guybedford:
Looks good. No strong feelings either way on error message testing, feel free to merge.

PR #76, opened 2024-07-11
fix: wasm32-wasi install for toolchain

guybedford:
We had a failed publish on Fastly due to a different toolchain being installed by default and not having the `wasm32-wasi` target loaded.

This adds an explicit `--toolchain` flag to the `rustup target add` command, which should avoid issues like that in future.


tschneidereit (approved):
Makes sense!

PR #80, opened 2024-07-22
fix: headers refactoring and state transition invariants

guybedford:
This includes some simple refactoring including unused code paths and also adds some extra state assertions, including encoding the state transition invariant that we never transition from `Uninitialized` to `HostOnly`. If this invariant changes in future we can always relax it again, but it helped my understanding of the state machine to note these assertions.

### DIFF
--- a/builtins/web/fetch/headers.cpp
+++ b/builtins/web/fetch/headers.cpp
@@ -530,7 +530,6 @@ JSObject *Headers::create(JSContext *cx, host_api::HttpHeadersReadOnly *handle, 
     return nullptr;
   }
 
-  MOZ_ASSERT(Headers::is_instance(self));
   MOZ_ASSERT(Headers::mode(self) == Headers::Mode::Uninitialized);
   SetReservedSlot(self, static_cast<uint32_t>(Headers::Slots::Mode),
                   JS::Int32Value(static_cast<int32_t>(Headers::Mode::HostOnly)));
--- a/builtins/web/fetch/headers.cpp
+++ b/builtins/web/fetch/headers.cpp
@@ -511,14 +511,6 @@ bool Headers::append_header_value(JSContext *cx, JS::HandleObject self, JS::Hand
   return true;
 }
 
-void init_from_handle(JSObject* self, host_api::HttpHeadersReadOnly* handle) {
-  MOZ_ASSERT(Headers::is_instance(self));
-  MOZ_ASSERT(Headers::mode(self) == Headers::Mode::Uninitialized);
-  SetReservedSlot(self, static_cast<uint32_t>(Headers::Slots::Mode),
-                  JS::Int32Value(static_cast<int32_t>(Headers::Mode::HostOnly)));
-  SetReservedSlot(self, static_cast<uint32_t>(Headers::Slots::Handle), PrivateValue(handle));
-}
-
 JSObject *Headers::create(JSContext *cx, host_api::HttpHeadersGuard guard) {
   JSObject* self = JS_NewObjectWithGivenProto(cx, &class_, proto_obj);
   if (!self) {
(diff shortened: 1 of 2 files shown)


builtins/web/fetch/headers.h
```diff
@@ -111,7 +111,7 @@ class Headers final : public BuiltinImpl<Headers> {
   static JSObject *create(JSContext *cx, host_api::HttpHeadersReadOnly *handle,
                           host_api::HttpHeadersGuard guard);
 
-  static JSObject *init_entries(JSContext *cx, HandleObject self, HandleValue init_headers);
+  static bool init_entries(JSContext *cx, HandleObject self, HandleValue init_headers);
```
JakeChampion, line 114: We need to make a decision at some point about how we version starlingmonkey and how we communicate public function signatures changes such as this

JakeChampion (approved):

builtins/web/fetch/headers.cpp
```diff
@@ -555,24 +543,28 @@ JSObject *Headers::create(JSContext *cx, HandleValue init_headers, host_api::Htt
   if (!self) {
     return nullptr;
   }
-  return init_entries(cx, self, init_headers);
+  if (!init_entries(cx, self, init_headers)) {
+    return nullptr;
+  }
+  MOZ_ASSERT(mode(self) == Headers::Mode::ContentOnly || mode(self) == Headers::Mode::Uninitialized);
```
tschneidereit, line 549: Right now, I'm pretty sure `mode(self)` will always be `Unitialized` here, because `init_entries` doesn't change it. And I think here we need to do the same thing as for the other `create` overload and explicitly set the state. Otherwise, `switch_mode` will reset the `entries` map to an empty `Map` at the next opportunity.

Which I guess in turn means that we can simplify `switch_mode` even further, because we should now have a setup in which it can never encounter the `Unitialized` state.
```suggestion
  MOZ_ASSERT(Headers::mode(self) == Headers::Mode::Uninitialized);
  SetReservedSlot(self, static_cast<uint32_t>(Headers::Slots::Mode),
                  JS::Int32Value(static_cast<int32_t>(Headers::Mode::ContentOnly)));
```

tschneidereit (changes requested):
This is a good insight, thank you!

I think there's a slight change still to be made—see the comments in-line, but then it'll be good to go.

builtins/web/fetch/headers.h
guybedford: From js-compute-runtime's perspective, every StarlingMonkey commit is potentially a breaking change. There simply are no API guarantees.

But since these are changes on the js-compute-runtime -> StarlingMonkey interface, they are all manageable in theory.

Problems come into play if:

1. A StarlingMonkey change causes the corresponding js-compute-runtime upgrade to be an inhibitative amount of work.
2. A StarlingMonkey change causes a breaking change in js-compute-runtimes's public guest API usage for JS users.

I hope our implicit agreement is effectively that we could revert a change in the case of (1) or (2), but perhaps we should discuss this more with @tschneidereit.

builtins/web/fetch/headers.cpp
guybedford: In this variation of headers create, `headersInit` contains iterable header values to populate the initial headers. So when `init_entries` does that population, the state is changed to `ContentOnly` (through append_value calling prepare_entries_for_modification), unless the `headersInit` was empty in which case the headers remain in `Uninitialized`.

So the assertion remains correct and non-redundant.

tschneidereit: Oh, of course! This only goes to show that simplifying this code is a very good idea! 😬

builtins/web/fetch/headers.h
tschneidereit: I agree with this as a sentiment for now, yes. Eventually I would like to get to a more stable API, but I don't think the one we have right now is sufficiently well designed to declare it stable as-is.

tschneidereit (approved):

PR #82, opened 2024-07-24
Update openssl url to new location

JakeChampion:
it looks like openssl broke their links when they launched their new website? https://www.openssl.org/source/openssl-3.0.7.tar.gz 404s and is now at https://openssl.org/source/old/3.0/openssl-3.0.7.tar.gz


tschneidereit (approved):
Thank you for fixing this!

PR #89, opened 2024-07-26
Add CMake option to explicitly skip wasm-opt and use `wasm-tools strip` instead.

cfallin:
As per #8, it seems the main purpose of using wasm-opt was to reduce module size by stripping debug sections. The comment in the CMake file also indicates an intent to optimize the binary.

However, `wasm-opt -O3` is extremely expensive on large programs; on my 18-core, 36-thread workstation, it took several minutes to complete. This is very painful when part of a debug loop that requires release builds (e.g. for performance features).

This PR adds a `USE_WASM_OPT` CMake option, on by default to preserve existing behavior, that allows one to do `cmake
-DCMAKE_BUILD_TYPE=Release -DUSE_WASM_OPT=OFF` to get a release build without `wasm-opt`. In its place, this PR uses `wasm-tools strip -a`, which strips debug sections (and all other custom sections) without rewriting code.


tschneidereit (approved):
Makes sense to me.

Just in case it matters for your development process: `wasm-opt` is unfortunately critical for achieving good performance, not just to reduce file size: LLVM's wasm32 target leaves a lot of optimizations to `wasm-opt`, so it's not really optional to run for (actual) release artifacts.

tschneidereit:
Oh, and another note: several minutes seems excessively long. On my M2 MBP, running `wasm-opt` on `starling.wasm` takes maybe 20-30 seconds. Still long enough to justify this PR, but the difference makes me wonder if something's up with your `wasm-opt`?

cfallin:
> Oh, and another note: several minutes seems excessively long. On my M2 MBP, running wasm-opt on starling.wasm takes maybe 20-30 seconds. Still long enough to justify this PR, but the difference makes me wonder if something's up with your wasm-opt?

Weirdly it took 3m55s for me (and Binaryen 117, one release behind current, downloaded by CMake):

```
cfallin@cfallin-tower:~/work/starlingmonkey/build% time ../deps/cpm_cache/binaryen/9dd4ea32e81e22d3849b452b706a16035cc86563/binaryen/bin/wasm-opt -O3 --strip-debug --enable-bulk-memory starling.wasm -o o.wasm
 -O3 --strip-debug --enable-bulk-memory starling.wasm -o o.wasm  688.73s user 6905.28s system 3220% cpu 3:55.77 total
cfallin@cfallin-tower:~/work/starlingmonkey/build% ../deps/cpm_cache/binaryen/9dd4ea32e81e22d3849b452b706a16035cc86563/binaryen/bin/wasm-opt --version
wasm-opt version 117 (version_117)
```

That system time stands out, and combined with (i) user-time of ~11 minutes (so average CPU occupancy of ~3 cores) and (ii) observation in `top` that this was pegging 36 cores tells me that Binaryen probably has some weird locking/scaling bottleneck (futex syscalls under mutexes?). My fault for having a large Xeon workstation I guess?

PR #90, opened 2024-07-26
 SpiderMonkey CMake package: use alternate artifacts if specified.

cfallin:
When developing StarlingMonkey using an existing SpiderMonkey build with no SM changes, it is very convenient to download and use pre-built artifacts for SM from the CI on `spidermonkey-wasi-embedding`, and we do not want to lose this convenience or approachability.

However, when modifying SpiderMonkey itself, it is nearly impossible to do development with the current setup: the debug loop requires merging changes to `gecko-dev` with one PR, updating the hash in `spidermonkey-wasi-embedding` with another PR, getting both merged and waiting for the build in CI before StarlingMonkey can use the new engine.

Rather than require a local hack for such cases, this modifies the SpiderMonkey CMake module to recognize a `SPIDERMONKEY_BINARIES` environment variable, and use artifacts there instead. This allows the developer to point this to a local clone of `spidermonkey-wasi-embedding` with a modified `gecko-dev`, rebuilt with `rebuild.sh` as needed. The `README` is updated with details on how to use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,3 +160,55 @@ name of one of the directories. Currently, only an implementation in terms of [w
 (host-apis/wasi-0.2.0) is provided, and used by default.
 
 To provide a custom host API implementation, you can set `HOST_API` to the (absolute) path of a directory containing that implementation.
+
+## Developing Changes to SpiderMonkey
+
+StarlingMonkey uses SpiderMonkey as its underlying JS engine, and by default,
+downloads build artifacts from [a wrapper
+repository](https://github.com/bytecodealliance/spidermonkey-wasi-embedding)
+around [our local SpiderMonkey
+tree](https://github.com/bytecodealliance/gecko-dev). That wrapper repository
+contains a SpiderMonkey commit-hash in a file, and its CI jobs build the
+artifacts that StarlingMonkey downloads during its build.
+
+This flow is optimized for ease of development of StarlingMonkey, and avoiding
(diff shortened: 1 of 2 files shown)


tschneidereit (commented):
I would somewhat prefer using an env var for this instead of hard-coding a specific path to check. Would you mind looking at [how the WPT harness does this](https://github.com/bytecodealliance/StarlingMonkey/blob/cc351f231f6f884b157defde60656fd039cf5cbc/tests/wpt-harness/wpt.cmake#L5-L15) and seeing if that kind of setup works for you?

And if it does work, please also update the `README` to document this.

cfallin:
Sure thing, updated! This now takes a `SPIDERMONKEY_BINARIES` environment variable. I've added a section at the end of `README.md` describing the workflow for developing the engine itself too.

(cc @guybedford I think from #81: [this check](https://github.com/bytecodealliance/StarlingMonkey/blob/cc351f231f6f884b157defde60656fd039cf5cbc/tests/wpt-harness/wpt.cmake#L5-L15) that Till linked above is checking `DEFINED $ENV{VAR}`, while CMake seems to require `DEFINED ENV{VAR}` (at least, that is what worked for me in testing this PR). I don't know enough about the WPT plumbing to test whether the above is doing what you mean but it might be good to verify?)

guybedford (approved):

PR #92, opened 2024-07-27
Implement `Request.clone`

### LINKED ISSUE #84: Request.clone() regression
This feature needs to be reimplemented to support the associated WPT test at tests/wpt-harness/expectations/fetch/api/request/request-structure.any.js.json for "Request has clone method".

tschneidereit:
This is fairly straightforward by itself. I threw in a couple of additional commits that clean a few things up. They come with detailed commit messages describing what they're about. Let me know if you'd prefer splitting them off, though.

Fixes #84


guybedford (approved):
Looks great. Since there isn't WPT coverage for this, can we add an integration test? For example, porting the cases from https://github.com/fastly/js-compute-runtime/blob/main/integration-tests/js-compute/fixtures/app/src/request-clone.js.

JakeChampion:
> Looks great. Since there isn't WPT coverage for this, can we add an integration test? For example, porting the cases from https://github.com/fastly/js-compute-runtime/blob/main/integration-tests/js-compute/fixtures/app/src/request-clone.js.

Perhaps we need to update wpt?
https://wpt.fyi/results/fetch/api/request/request-clone.sub.html?label=experimental&label=master&aligned

guybedford:
@JakeChampion we are on a recent WPT, we just don't run `html` tests.

guybedford:
@tschneidereit if it helps, I added a simple clone test here, which is giving an error just on this line:

```
deepStrictEqual([...newRequest.headers], [...request.headers], 'newRequest.headers');
```

with the stdout when that line is enabled (from tests/integration/fetch/stderr.log) reading:

```
stderr [0] :: Warning: Using a DEBUG build. Expect things to be SLOW.
stderr [0] :: "POST"
stderr [0] :: [-1] Assertion failure: isObjectOrNull(), at 
stderr [0] :: /Users/gbedford/Projects/StarlingMonkey/deps/cpm_cache/spidermonkey-debug/76821ccb5a765a81c18400e00f9c7207f1fd9461/spidermonkey-debug/include/js/Value.h
stderr [0] :: :998
2024-07-30T00:10:41.598639Z ERROR wasmtime_cli::commands::serve: [0] :: Error {
```

guybedford:
I went ahead and pushed up a fix for the headers cloning bug, will land this later today.

tschneidereit:
Thank you for doing that! <3

Can you say what the fix actually was? That commit for me renders with lots of changes that seem to only change whitespace (even if I set the "ignore whitespace changes" flag) so it's hard to tell what actually changed

tschneidereit:
Oh, never mind: clicking on the commit in the email notification showed something entirely different from what I get when viewing only this commit in the PR itself 🤷🏻

guybedford:
Yes, I applied clang format by mistake, then pushed up the correction.

tschneidereit:
Ah, that makes sense!

As does the fix :)

PR #100, opened 2024-07-29
fix: providing a custom WPT_ROOT env var

guybedford:
As identified by @cfallin this fixes the `WPT_ROOT` environment variable to be correctly referenced.


cfallin:
FWIW, I had found that `DEFINED ENV{...}` without the sigil is needed for the conditional (otherwise I guess the actual value is subbed in?) but `$ENV{...}` with the sigil at use-sites was needed?

JakeChampion (approved):

PR #103, opened 2024-08-01
Add Rust bindings for the runtime and the SpiderMonkey engines

tschneidereit:
This PR introduces rust-bindgen generated bindings for the StarlingMonkey runtime itself, as well as generated bindings and higher-level abstractions building on those for the SpiderMonkey engine.

The latter are derived from the Servo project's [mozjs](https://github.com/servo/mozjs) crate and share much of the code. The build system is substantially overhauled though to enable direct use of the [Bytecode Alliance's fork](https://github.com/bytecodealliance/gecko-dev/) of the repository containing SpiderMonkey, instead of having to update tarball based source code drops.

To ensure compliance with mozjs's license and easy code movement between files, all Rust code in the core runtime is licensed under the MPL2 instead of the Bytecode Alliance's default Apache 2.0 with LLVM exceptions. At the board meeting on 2024-07-25, the BA's board approved a license exception to permit use of the MPL2 for part or all of the StarlingMonkey project, as required by our [IP policy](https://bytecodealliance.org/assets/ip-policy.pdf).


guybedford:
@andreiltd please do!

tschneidereit:
Indeed, please do!

I just pushed a couple more commits here, one of which I had lying around and that might be helpful, and the other containing documentation of the authorship situation of the Rust bindings. The latter is crucial for landing any of this.

One thing to note is that because of some issues with the bindings generation, the bindings will sometimes be [wrong in subtle ways](https://github.com/rust-lang/rust-bindgen/issues/2871). There's not a huge amount we can do about that, but it might be useful to at least document somehow.

PR #104, opened 2024-08-02
Update upstream SpiderMonkey for a PBL+weval fix.

cfallin:
Pulls in bytecodealliance/gecko-dev#56.


cfallin:
Staging this along with bytecodealliance/spidermonkey-wasi-embedding#20 in parallel, both of which depend on bytecodealliance/gecko-dev#56; will collapse the dominoes in series then propagate downstram.

JakeChampion (approved):

Issue #101, opened 2024-07-31
URL Finalizer


guybedford:
In the `URL` builtin, we associate all URLs with a Rust URL implementation in https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/url.cpp#L617, stored in a private slot.

This pointer is stored as a private value, which according to the SpiderMonkey docs sounds like it is never marked by the GC (https://searchfox.org/mozilla-central/source/js/public/Value.h#1052).

So it isn't clear to me how we are releasing these URL allocations. Shouldn't there be some kind of GC hook required to achieve that here, to make it known to SpiderMonkey that these are pointers to classes that must be disposed?

PR #105, opened 2024-08-02
gc: add finalization for URL

guybedford:
This is a first attempt to fix https://github.com/bytecodealliance/StarlingMonkey/issues/101 for further feedback.

In testing, I'm still not able to get the finalizer to run, as I believe we disable the GC during runtime still.

That said, supporting the GC running for long-running executions, and coding to avoid memory leaks still seem like important properties to retain for the runtime in general, even if not used.

@tschneidereit would very much value your feedback further here.

### DIFF
--- a/builtins/web/url.cpp
+++ b/builtins/web/url.cpp
@@ -580,6 +580,11 @@ JSObject *URL::create(JSContext *cx, JS::HandleObject self, JS::HandleValue url_
   return create(cx, self, url_val, base);
 }
 
+void URL::finalize(JS::GCContext *gcx, JSObject *self) {
+  jsurl::JSUrl* url = static_cast<jsurl::JSUrl *>(JS::GetReservedSlot(self, Slots::Url).toPrivate());
+  free(url);
+}
+
 JSObject *URL::create(JSContext *cx, JS::HandleObject self, JS::HandleValue url_val,
                       JS::HandleValue base_val) {
   if (is_instance(base_val)) {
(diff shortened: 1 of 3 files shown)


tschneidereit (commented):
It definitely makes sense to add the ability to free natively allocated data in builtins, and the `JSUrl` specifically.

The thing I'm a bit worried about is how this makes every builtin object have a finalizer: I remember that that introduces some amount of overhead, which we might not want to take on lightly.

It's entirely possible that this doesn't currently apply, because it might've mainly been about forced foreground finalization. But even then, I think we should try to design things such that any eventual multithreading wouldn't be substantially slowed down.

Given all this, two things might make sense:
1. Introduce a `FinalizableBuiltinImpl` extending `BuiltinImpl` and only having the finalizer on that
2. setting `JSCLASS_BACKGROUND_FINALIZE` instead of `JSCLASS_FOREGROUND_FINALIZE` (or both: I don't at all remember how these work, and whether they're additive or exclusive)

tschneidereit:
Oh also, I think we should introduce a testing builtin that supports forcing GCs and such to be able to test this. Probably with an API that's compatible with what some of the WPT tests check for.

guybedford:
Thanks for the thoughtful feedback here, very much agreed on both.

guybedford:
I've updated this PR to use background finalization and a new `FinalizableBuiltinImpl` as described.

I still haven't implemented a full `window.gc()` builtin (per Node.js) or otherwise for GC testing though.

Since this is currently strictly speaking a bug, shall we land this and go with this approach for now?

tschneidereit (approved):
This makes a lot of sense to me, and looks good!

(I didn't check if the unrelated formatting changes are all correct, so might be good to double check before merging.)

tschneidereit:
Also, agreed on landing this now without blocking on the `gc()` builtin.

guybedford:
Excellent, thanks, the formatting is just clang format.

Issue #85, opened 2024-07-24
Non-ascii bytes in request headers


guybedford:
WPT test regression from tests/wpt-harness/expectations/fetch/api/basic/request-headers-nonascii.any.js.json.

guybedford:
This was resolved in #108.

Issue #78, opened 2024-07-18
Implement "immutable" headers guard


guybedford:
Fetch requests should return a response with headers marked with an `"immutable"` guard, where `check_guard` should always fail. In the Fetch spec it appears this is only ever done for responses.

Would this make sense to consider as an addition? Or is another implementation mechanism planned?

guybedford:
Or is this functionality fully covered by the `HttpHeadersReadOnly` handle?

tschneidereit:
It is indeed fully covered by `HttpHeadersReadOnly`, yes. I'm not opposed to additionally use the guard concept for this, though that'd obviously cause duplication.

I'd definitely prefer to not replace the type system enforcement of read-only-ness with a dynamic check completely though: introducing the type hierarchy has definitely helped me fix various bugs.

guybedford:
Resolved in #108.

PR #108, opened 2024-08-08
fix: headers WPT conformance

guybedford:
This fixes 40 web platform tests, without any WPT regressions, related to the headers subsystem including header encoding, validation, immutable guards, header ordering, multi-value sorting, iteration.

In addition it implements the `Headers.prototype.getSetCookie` method, resolving all tests for it.

Resolves:

* Implement headers `getSetCookie`: https://github.com/bytecodealliance/StarlingMonkey/issues/109
* Non-ascii bytes in request headers (request-headers-nonascii.any.js.json): https://github.com/bytecodealliance/StarlingMonkey/issues/85
* Immutable headers guards for incoming response headers (response-headers-guard.any.js.json) : https://github.com/bytecodealliance/StarlingMonkey/issues/78

Headers in the `CachedInContent` and `ContentOnly` state are stored as an ordered vector of `host_api::HostString` to match the representation on the wire, without the need for reencoding across the host API calls.

A separate `sort_list` is maintained separately representing the ordered iteration through the headers, and iterator and getter functions are updated to use this list when appropriate.

In addition header keys are stored with their original casing to match WhatWG, with equality and ordering defined using case-invariant comparison operations.

Header casing tests on the wire are not passing since Wasmtime will lowercase headers, but if the WASI implementation were to support header casing then this would work (eg js-compute-runtime and Jco may choose to support header casing under this implementation).

Guard validations need to happen on every append so the guard logic is moved back into the main headers implementation, with new host calls `get_forbidden_request_headers()` and `get_forbidden_response_headers()` used to obtain the information for the header guards.


tests/integration/tests/performance.js
```diff
@@ -1,285 +0,0 @@
-import { serveTest } from "../test-server.js";
```
JakeChampion, line 1: I believe this was deleted because a duplicate of `tests/integration/performance/performance.js`

builtins/web/fetch/headers.h
```diff
@@ -53,34 +54,61 @@ class Headers final : public BuiltinImpl<Headers> {
...
-  static bool set_if_undefined(JSContext *cx, JS::HandleObject self, string_view name,
-                               string_view value);
+  enum class HeadersGuard {
+    None,
+    Request,
+    Response,
+    Immutable,
+  };
+
+  /// Adds the valid given header name/value to `self`'s list of headers iff `self`
```
JakeChampion, line 89:
```suggestion
  /// Adds the valid given header name/value to `self`'s list of headers if `self`
```

builtins/web/fetch/request-response.cpp
```diff
@@ -497,9 +495,11 @@ bool RequestOrResponse::append_body(JSContext *cx, JS::HandleObject self, JS::Ha
 JSObject *RequestOrResponse::headers(JSContext *cx, JS::HandleObject obj) {
   JSObject *headers = maybe_headers(obj);
   if (!headers) {
-    host_api::HttpHeadersGuard guard = Request::is_instance(obj)
-      ? host_api::HttpHeadersGuard::Request
-      : host_api::HttpHeadersGuard::Response;
+    // Incoming request and incoming response headers are immutable per service worker
```
JakeChampion, line 498: I understand this is making the implementation spec compliant which is good, but this is also a breaking change as some applications will now throw an error because of this change. With that in my, I think this should be released as a new major version (if not in StarlingMonkey then in downstream projects)

builtins/web/fetch/headers.h
tschneidereit: The "iff" here stands for "[if, and only if](https://en.wikipedia.org/wiki/If_and_only_if)", so it's not a spelling mistake :)

tests/integration/tests/performance.js
guybedford: Yes exactly.

builtins/web/fetch/request-response.cpp
guybedford: Since this is implemented in the `request-response` usage of Headers, which Fastly maintains its own version of, we can choose to not implement this yet in js-compute-runtime to maintain this as a non-breaking change.

I'll make sure to include this in the Fastly PR at https://github.com/fastly/js-compute-runtime/pull/844.

builtins/web/fetch/headers.cpp
```diff
@@ -923,12 +1079,188 @@ unique_ptr<host_api::HttpHeaders> Headers::handle_clone(JSContext* cx, HandleObj
...
+    return true;
+  }
+
+  JS_DefineProperty(cx, result, "done", JS::FalseHandleValue, JSPROP_ENUMERATE);
+
+  JS::RootedValue key_val(cx);
+  JS::RootedValue val_val(cx);
+
+  // If there was a header prepend, we must recompute the distinct index for proper iteration.
+  // We detect prepends by checking the last_key sorting.
```
guybedford, line 1203: Writing this comment for myself - this comment is out of date from a former version of this code.

builtins/web/fetch/headers.cpp
```diff
@@ -923,12 +1080,181 @@ unique_ptr<host_api::HttpHeaders> Headers::handle_clone(JSContext* cx, HandleObj
...
+  // `constructor` property on `HeadersIterator.prototype`. The latter
+  // because Iterators don't have their own constructor on the prototype.
+  return JS_DeleteProperty(cx, global, class_.name) &&
+         JS_DeleteProperty(cx, proto_obj, "constructor");
+}
+
+std::tuple<host_api::HostString, host_api::HostString> *
+Headers::get_index(JSContext *cx, JS::HandleObject self, size_t index) {
+  MOZ_ASSERT(is_instance(self));
+  std::vector<size_t> *headers_sort_list = static_cast<std::vector<size_t> *>(
```
tschneidereit, line 1141: For both `HeadersList` and `HeadersSortList` it'd be very nice to introduce getter functions to eliminate the noise of retrieving them here and in a bunch of other places.

E.g. for `HeadersSortList`, I'd imagine that to be
```c++
Headers::headers_sort_list(JSObject* self) {
  MOZ_ASSERT(is_instance(self));
  auto *headers_sort_list = static_cast<std::vector<size_t> *>(
      JS::GetReservedSlot(self, static_cast<size_t>(Headers::Slots::HeadersSortList)).toPrivate());
  MOZ_ASSERT(headers_sort_list);
  return headers_sort_list;
}
```

reducing the first four lines of this function here to just
```c++
  auto *headers_sort_list = headers_sort_list(self);
```

builtins/web/fetch/headers.cpp
```diff
@@ -923,12 +1080,181 @@ unique_ptr<host_api::HttpHeaders> Headers::handle_clone(JSContext* cx, HandleObj
...
+      // headers should already be validated by here
+      MOZ_ASSERT(ch <= 127 && VALID_NAME_CHARS[ch]);
+      // we store header keys with casing, so getter itself lowercases
+      if (ch >= 'A' && ch <= 'Z') {
+        chars[i] = ch - 'A' + 'a';
+      } else {
+        chars[i] = ch;
+      }
+    }
+    key_val = JS::StringValue(JS_NewLatin1String(cx, JS::UniqueLatin1Chars(chars), len));
```
tschneidereit, line 1218: This needs an OOM check for the `JS_NewLatin1String` before creating the `StringValue`

host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -357,88 +344,50 @@ Result<HttpHeaders*> HttpHeaders::FromEntries(HttpHeadersGuard guard,
...
-
-  if (!forbidden_headers) {
-    return true;
-  }
-
-  for (auto header : *forbidden_headers) {
-    if (header_name.compare(header) == 0) {
-      return false;
-    }
+std::vector<string_view> HttpHeaders::get_forbidden_request_headers() {
```
tschneidereit, line 372: Would this be better as a `const&` to ensure the vec itself isn't copied, while it's also not mutated?

tschneidereit (changes requested):
This is absolutely great, thank you for all the effort you put into it! <3

I left a whole bunch of comments, but I don't think addressing them should be too hard, and none of them point out major shortcomings of any kind.

I'd still like to take one more look, so please don't force-push changes so I can do an incremental diff.

builtins/web/fetch/headers.cpp
```diff
@@ -537,24 +602,24 @@ JSObject *Headers::create(JSContext *cx, host_api::HttpHeadersReadOnly *handle,
...
     return nullptr;
   }
-  MOZ_ASSERT(mode(self) == Headers::Mode::ContentOnly || mode(self) == Headers::Mode::Uninitialized);
+  MOZ_ASSERT(mode(self) == Headers::Mode::ContentOnly ||
+             mode(self) == Headers::Mode::Uninitialized);
   return self;
 }
 
 bool Headers::init_entries(JSContext *cx, HandleObject self, HandleValue initv) {
-  // TODO: check if initv is a Headers instance and clone its handle if so.
```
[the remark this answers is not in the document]
guybedford: Sure, readded - but just to note that Fastly doesn't have the capability to do headers handle cloning, so this would ideally be implemented in a way that can be turned off.

builtins/web/fetch/headers.cpp
```diff
@@ -622,23 +720,48 @@ bool Headers::set(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+      HeadersList *headers_list = static_cast<Headers::HeadersList *>(
+          JS::GetReservedSlot(self, static_cast<size_t>(Headers::Slots::HeadersList)).toPrivate());
+      MOZ_ASSERT(headers_list);
+
+      // Update the first entry in place to the new value
+      host_api::HostString *header_val =
+          &std::get<1>(headers_list->at(headers_sort_list->at(index)));
+
+      // Swap in the new value respecting the disposal semantics
+      header_val->ptr.swap(value_chars.ptr);
```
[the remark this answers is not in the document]
guybedford: Because `header_val` is a pointer to a host string, equality doesn't work.

builtins/web/fetch/headers.cpp
```diff
@@ -622,23 +720,48 @@ bool Headers::set(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+             (header_compare(std::get<0>(headers_list->at(
+                                 headers_sort_list->at(index + delete_cnt + 1) - delete_cnt)),
+                             name_chars) == Ordering::Equal)) {
+        headers_list->erase(headers_list->begin() + headers_sort_list->at(index + delete_cnt + 1) -
+                            delete_cnt);
+        delete_cnt++;
+      }
+      // Reset the sort list if we performed additional deletions.
+      if (delete_cnt > 0) {
+        headers_sort_list->clear();
```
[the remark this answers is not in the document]
guybedford: Yes, this is possible with some careful reindexing work, not sure I have the time to do it now though.

builtins/web/fetch/headers.cpp
```diff
@@ -622,23 +720,48 @@ bool Headers::set(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+      // Delete all subsequent entries for this header excluding the first,
+      // as a variation of Headers::delete.
+      size_t len = headers_list->size();
+      size_t delete_cnt = 0;
+      while (index + delete_cnt + 1 < len &&
+             headers_sort_list->at(index + delete_cnt + 1) >= delete_cnt &&
+             (header_compare(std::get<0>(headers_list->at(
+                                 headers_sort_list->at(index + delete_cnt + 1) - delete_cnt)),
+                             name_chars) == Ordering::Equal)) {
+        headers_list->erase(headers_list->begin() + headers_sort_list->at(index + delete_cnt + 1) -
```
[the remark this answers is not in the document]
guybedford: This isn't possible because we don't assume that same-value headers are contiguous.

builtins/web/fetch/headers.cpp
```diff
@@ -760,81 +912,66 @@ bool Headers::delete_(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+          JS::GetReservedSlot(self, static_cast<size_t>(Headers::Slots::HeadersList)).toPrivate());
+      MOZ_ASSERT(headers_list);
+
+      // Delete all case-insensitively equal names.
+      // The ordering guarantee for sort_list is that equal names will come later in headers_list
+      // so that we can continue to use sort list during the delete operation, only recomputing it
+      // after.
+      size_t delete_cnt = 0;
+      size_t len = headers_sort_list->size();
+      do {
```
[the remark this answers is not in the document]
guybedford: As above, this requires the contiguous assumption for headers_list.

builtins/web/fetch/headers.cpp
guybedford: `JS_NewLatin1String` takes ownership of the unique chars, and so there is no memory allocation or failure associated.

host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -357,88 +344,50 @@ Result<HttpHeaders*> HttpHeaders::FromEntries(HttpHeadersGuard guard,
...
+  std::vector<string_view> list;
+  for (auto it = forbidden_response_headers.begin(); it != forbidden_response_headers.end(); ++it) {
+    list.push_back(*it);
   }
-
-  return headers;
+  return list;
 }
 
+// TODO: actually use WASI fields->clone() to avoid mutating the input!
```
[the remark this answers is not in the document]
guybedford: Yes you're right thanks, again this is a Fastly-specific limitation, where we need to clearly constrain the model to avoid cloning.

host-apis/wasi-0.2.0/host_api.cpp
guybedford: It is mean't to be mutated so we can take ownership of this list and then ensure it is properly sorted.

So the only change would be just be moving the clone into the caller, which I'm not sure is necessary?

guybedford:
Thanks for the positive reception and detailed feedback here, I've resolved everything that's been addressed, leaving the outstanding comments open. The `pr feedback` commit contains everything in that diff. I may push some subsequent changes as additional comments and will comment about what those are if I do them.

builtins/web/fetch/headers.cpp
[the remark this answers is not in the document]
guybedford: We could assume same-value headers are contiguous if we always perform joining when populating the headers list from the host. I only discovered late in the implementation that this is truly unobservable since passing headers to a response always runs through a re-iteration and sorting.

The observability of different orderings is then only observable in the case of `new Response().headers.append('unordered')` allowing unordered construction on the wire.

Note that this is also a justification for treating `initv` as always running through the guest iteration, since joining and sorting is an expectation here, whereas a host clone in theory would not join and sort.

tschneidereit (approved):
This is great, thank you! A few more comments, but nothing that holds up landing

host-apis/wasi-0.2.0/host_api.cpp
guybedford: Added something along these lines.

builtins/web/url.cpp
```diff
@@ -187,8 +186,7 @@ jsurl::SpecSlice URLSearchParams::serialize(JSContext *cx, JS::HandleObject self
 
 bool URLSearchParams::append(JSContext *cx, unsigned argc, JS::Value *vp) {
   METHOD_HEADER(2)
-  bool err = false;
-  auto value = append_impl_validate(cx, args[0], &err, "append");
+  auto value = append_impl_validate(cx, args[0], "append");
```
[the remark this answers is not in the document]
guybedford: The URL iteration doesn't do key validation before value validation like the headers one does currently, although further investigation into the failing URL tests might indicate implementing this can resolve some remaining WPT there. As a headers PR I didn't want to get into that here though.

Issue #112, opened 2024-08-16
Transform stream `readable_get` and `writable_get` assertion failures


guybedford:
In the WPT debug build, `readable_get` and `writable_get` call `readable` and `writable` respectively, which fail an assertion when looking like:

```
JSObject *TransformStream::readable(JSObject *self) {
  MOZ_ASSERT(is_instance(self));
  return &JS::GetReservedSlot(self, TransformStream::Slots::Readable).toObject();
}
```

that the slot fails `toObject()`.

PR #113, opened 2024-08-16
ci: add WPT debug build, fix streams assertion failure

guybedford:
This adds a CI job for the debug build on WPT, and also fixes an assertion failure which looks like a bug in transform streams.

I've posted a separate issue for this bug in https://github.com/bytecodealliance/StarlingMonkey/issues/112 to be resolved separately.

Necessary for Fastly's WPT to pass for the debug build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,7 +76,7 @@ jobs:
     name: Web Platform Tests
     strategy:
       matrix:
-        build: [release]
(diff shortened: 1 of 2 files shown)


cfallin (approved):
LGTM, this is at least an incremental improvement; thanks for filing the tracking issue as well and hopefully we can work out the root-cause eventually.

PR #114, opened 2024-08-16
ci: Weval on WPT matrix

guybedford:
This extends Weval into the WPT CI matrix, also adding it into a matrix for the main test run as well.


.github/workflows/main.yml
```diff
@@ -94,7 +69,9 @@ jobs:
 
     - name: Build StarlingMonkey WPT
       run: |
-        cmake -S . -B cmake-build-${{ matrix.build }} -DENABLE_WPT:BOOL=ON -DCMAKE_BUILD_TYPE=${{ matrix.build == 'release' && 'Release' || 'Debug' }}
+        cmake -S . -B cmake-build-${{ matrix.build }} -DENABLE_WPT:BOOL=ON\
+          -DCMAKE_BUILD_TYPE=${{ (matrix.build == 'release' || matrix.build == 'weval') && 'Release' || 'Debug' }}\
+          ${{matrix.build == 'weval' && '-DUSE_WASM_OPT=OFF -DWEVAL=ON' || ''}}
```
cfallin, line 74: I think the failure here (a missing "database file") is the wevalcache file not being downloaded; it's a separate artifact IIRC?

cfallin (approved):
LGTM, I think the CI failure is just a file copy/plumbing issue hopefully...

.github/workflows/main.yml
guybedford: This workflow does the build and test run in the same task, so it should be there.

I think the issue is we need the WPT plumbing here for the WPT build script like in js-compute-runtime to run Weval componentize properly?

cfallin: I've got to run just now but I was able to get this to build locally by building the `starling-ics.wevalcache` target first: `cmake --build cmake-build --parallel 4 starling-ics.wevalcache` then `cmake ... wpt-runtime`. Perhaps there's a way to get the dependencies right to handle that automatically, or otherwise we can just add the extra target here?

guybedford:
@cfallin there seem to be 4 missing / failing tests here, which might be worth looking into as possible Weval bugs.

guybedford:
Seems to be in `fetch/api/headers/header-values.any.js`.

guybedford:
@cfallin that is an expected failure - the case here is specifically a panic for `header-values.any.js` resulting in the log `MISSING TEST...`. I can help dig into this further if you need.

cfallin:
It appears to be an issue of memory size limits:

```
wasmtime stderr: Error: memory index 0 has a minimum page size of 165 which exceeds the limit of 160
```

Now working out exactly where the limit is set and how to update this for the `wasmtime serve` command we run for WPT.

cfallin:
Actually, that appears to be an issue with an old Wasmtime ([Zulip thread](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/.E2.9C.94.20wasmtime.20serve.20wasm.20limits.20settings)); upgrading my local Wasmtime and running locally, I am unable to reproduce when running just the one test:

```
% node tests/wpt-harness/run-wpt.mjs -vv header-values.any.js
Starting WPT server (cmd: /home/cfallin/work/starlingmonkey/tests/wpt-harness/wpt/wpt --no-h2 serve)...
Starting Wasmtime server (cmd: wasmtime serve -W max-memory-size=134217728 -S common --addr 127.0.0.1:7676 wpt-runtime.wasm)...
wasmtime stderr: Serving HTTP on http://127.0.0.1:7676/
[ ... ]
fetch/api/headers/header-values.any.js    3 /    4 (   +0,    -0,    ?0) passing in    0ms

Done. Stats:                              3 /    4 (   +0,    -0,    ?0) passing in    0ms
```

but I see the failure when running the whole suite. I'm not sure where to go from here -- "event loop error - both task and job queues are empty, but expected operations did not resolve" does not tell me much, and without a single testcase I can try to reduce further I'm not really able to deduce anything. @guybedford are you by chance able to help derive a smaller testcase or do you have any ideas here?

guybedford:
@cfallin I updated Wasmtime but that doesn't seem to be the issue. I was able to verify again it is definitely the headers test though. Happy to run through it with you further when you have a moment.

PR #130, opened 2024-08-27
add links to consuming projects in readme

guybedford:
This updates the readme to note projects which are using StarlingMonkey (expected to be a list that can grow over time), which also seems useful for users to easily find where to use StarlingMonkey without having to build it themselves.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,0 +1,9 @@
+# StarlingMonkey Adopters
+
+_If you are using StarlingMonkey in production at your organization, please add your company name to this list.
+The list is in alphabetical order._
+
+| Organization | Contact | Status | Description of Use |
+| - | - | - | - |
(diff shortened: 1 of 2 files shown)


tschneidereit (commented):
Adding this makes a lot of sense, yes. I wonder though if we shouldn't introduce an `ADOPTERS` file like other projects have? E.g. Wasmtime [has such a file](https://github.com/bytecodealliance/wasmtime/blob/main/ADOPTERS.md), with a structure I quite like.

We could then include a link to that file as part of a single-sentence paragraph here.

guybedford:
Nice, one counter argument is that I did not know that page existed until you mentioned it. I really like the format though and table, but I think highlighting production users at the forefront of the readme can help build confidence in the project.

tschneidereit:
I agree, yes. That's what I was trying to address with the suggestion of including a mention/link in the README, but maybe inlining it is still the better call. In that case, maybe we do the inverse: introduce an ADOPTERS file that just contains a sentence with a link to this section of the README, so people expecting that kind of pattern immediately find what they're looking for?

guybedford:
Added a separate file and short summary, let me know your thoughts.

tschneidereit (approved):
I like it!

PR #132, opened 2024-08-28
feat: better errors with testing, dump_error() function

guybedford:
This improves the error reporting as syntax errors still display currently as `(new SyntaxError('err', 'file.js', 4))`.

With this change a proper reporter output is generated for pending exceptions.

In addition a new `dump_error()` function is exposed, which I've found to be useful for ComponentizeJS to also be able to report nicer errors.


elliottt (approved):
Looks good to me!

PR #135, opened 2024-08-28
use consistent header naming scheme

guybedford:
Updates the header definition names to use a consistent scheme corresponding to their folder structures.


elliottt (approved):
:shipit:

PR #139, opened 2024-08-30
Update event loop stall handling

guybedford:
This updates the handle of event loop stalls to update the current log message output of:

```
stderr [0] :: event loop error - both task and job queues are empty, but expected operations did not resolve
stderr [0] :: Internal error.
```

To now read:

```
stderr [0] :: Warning: JS event loop terminated without completing the request.
```

and includes a test for this behaviour.

In addition the error message printing is moved into the caller of the event loop instead of the callee to allow other embeddings to customize this message further.

This is based to and should land after https://github.com/bytecodealliance/StarlingMonkey/pull/132.


elliottt (approved):
Looks good! I read the last two commits, as the first at the time of review was from #132.

PR #143, opened 2024-09-10
fix: panic on transform stream check

guybedford:
This fixes a panic Fastly was seeing when passing a request body to a response, where this slot was not a boolean causing a panic.


cfallin (approved):
This seems locally fine as a more robust check, but do we know why the slot was not a boolean? Is there a path that doesn't initialize it, or is some other value being stored there erroneously?

guybedford:
There's a lot of complex logic here and quite honestly I haven't traced it carefully, but I can verify this solves the full end to end Fastly issue in https://github.com/fastly/js-compute-runtime/pull/954.

PR #146, opened 2024-09-24
fix: keep names section under RelWithDebInfo build

guybedford:
This ensures that `wasm-opt` does not strip the `"name"` section under `RelWithDebugInfo` builds (needed for ComponentizeJS).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if (USE_WASM_OPT)
         add_custom_command(
                 TARGET starling.wasm
                 POST_BUILD
-                COMMAND ${WASM_OPT} --strip-debug -O3 -o starling.wasm starling.wasm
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ if (USE_WASM_OPT)
         add_custom_command(
                 TARGET starling.wasm
                 POST_BUILD
-                COMMAND ${WASM_OPT} --strip-debug -O3 -o starling.wasm starling.wasm
+                COMMAND ${WASM_OPT} --strip-debug -O3 -g -o starling.wasm starling.wasm
```
cfallin, line 70: Title says `RelWithDebInfo` but this is the `Release` case I think -- should this be below instead?

cfallin (approved):

PR #147, opened 2024-09-24
Improve runtime configuration and initialization

tschneidereit:
This PR introduces a (somewhat rudimentary) options and arguments parser for StarlingMonkey and uses it to apply runtime configuration both when wizening and when invoked via `wasi:cli/run/run` or `wasi:http/incoming-handler/handle` without previous wizening.

Additionally, the runtime (called `api::Engine`, which we should probably change) now has a slightly more elaborate initialization system with multiple distinct states.

I also updated the version of wit-bindgen and WASI bindings in a separate PR, which should need a review only of the wit-bindgen update itself, not the generated bindings.


guybedford (commented):
My main feedback on this approach would be that we are heavily tying the engine code to the args parser.

It could be nice to treat the Engine as a high-level API, that the args parser uses - ie `Engine::enable_debug(bool)`, `Engine::enable_verbose(bool)`, `Engine::enable_module_mode(bool), `Engine::set_content_script(string)`, `Engine::eval(string)` (all with state machine guards to ensure they are called when they make sense).

And then to have a new Runtime::initialize(args) that acts as the CLI entry point calling to the engine, just so there's a little separation.

Projects building on StarlingMonkey can then choose to use the arg parser (which can also run in the CLI mode by default under a main model), or they could just call the Engine APIs directly, which ComponentizeJS probably wants to do. Currently we create our own Wizer init function, then call the StarlingMonkey Wizer init function from that, but it would be nice if we could fully write our own init by effectively constructing an equivalent set of engine initialization calls and then not needing to call the StarlingMonkey wizer entry at all and have our own initialization system.

Just some initial thoughts on this, not requesting changes per se.

tschneidereit:
@guybedford thank you for the feedback, much appreciated! I've now reworked things to extract arg parsing into its own header file. The engine itself only consumes a struct with arguments, which can be populated in whatever way works best for the embedder. (That was the case in the previous version as well, but that was far from obvious.)

What hasn't changed is that the engine is initialized based on the arguments, including the attempt to load a script if provided. The current behavior stays the same, by-and-large, though: if no script is provided, nothing happens.

guybedford (approved):

PR #148, opened 2024-09-27
fix: implement the finalizer pattern for the Headers builtin

guybedford:
This implements the GC finalization for the `Headers` builtin, disposing the private vectors correctly.


tschneidereit (approved):
Thank you!

PR #151, opened 2024-10-07
fix: componentize script to use local bins

guybedford:
This fixes the `componentize.sh` script to instead reference the locally installed toolchain, removing the need to have Wizer and wasm-tools in path.


cfallin (approved):
LGTM! I should've carried over the approach I took with weval to the other tools; thanks for finishing it out :-)

PR #161, opened 2024-10-16
Make starling.wasm a usable component, and document it as such

tschneidereit:
This changes the current `starling.wasm` build target to instead be called `starling-raw.wasm` to reflect that it's not yet a usable thing. Additionally, the current `runtime-eval` target is renamed to `starling` and produces a new `starling.wasm` file, which this time is a component that can be used to run dynamically loaded code.

I also updated the documentation accordingly.

And finally, I fixed a bug in `CMakeLists.txt` introduced in #151 that broke the `runtime-eval` build target. That's the entire reason I fell down this rabbit hole 😬

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ if (ENABLE_WPT)
     include("tests/wpt-harness/wpt.cmake")
 endif()
 
-add_executable(starling.wasm
+add_executable(starling-raw.wasm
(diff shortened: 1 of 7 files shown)


guybedford (approved):
Nice, I wasn't aware of this use case thanks for documenting it. Would be great to get this into CI as a follow-up to ensure we don't regress again.

PR #163, opened 2024-10-18
Fix generation of Ninja files

andreiltd:
Currently, when trying to build the project using the Ninja generator instead of Makefile, the build fails with the following error:

```
cmake -S . -B cmake-build-debug -DCMAKE_BUILD_TYPE=Debug -G Ninja
cmake --build cmake-build-debug

ninja: error: build.ninja:997: bad $-escape (literal $ must be written as $$)
```

This patch fixes the issue by checking for the actual make program name specifically for the Ninja generator.


guybedford (approved):
If it's working for you and passing to CI I'm fine to land this.

PR #166, opened 2024-10-22
fix: ensure filename is defined for builtins in debug builds

guybedford:
This fixes a bug where `filename` on CompileOptions was not being set for builtin modules in the script loader, which was fine on release builds but failing on debug builds.


tschneidereit (approved):
Thank you, and sorry for not catching that. I assume a test isn't trivial to write for this?

Issue #121, opened 2024-08-27
Blob and Request.blob


guybedford:

andreiltd:
I will have a look.

Issue #126, opened 2024-08-27
Readable Stream Methods


guybedford:
* ReadableByteStream tee
* Async iteration / BYOB async iteration
* ReadableStream.from ([https://github.com/whatwg/streams/pull/1083](https://github.com/whatwg/streams/issues/1018))

PR #169, opened 2024-10-23
Add Blob builtin implementation

andreiltd:
This patch adds support for the [Blob API](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

The current status of platform tests is ~~205~~ 216 passes out of 222 tests. Notable failures include:

- Constructor with DOM objects,
- BYOB stream support


tests/wpt-harness/tests.json
```diff
@@ -163,6 +163,18 @@
...
+  "FileApi/blob/Blob-bytes.any.js",
+  "FileApi/blob/Blob-constructor.any.js",
+  "FileApi/blob/Blob-constructor-dom.window.js",
+  "FileApi/blob/Blob-constructor-endings.html",
+  "FileApi/blob/Blob-in-worker.worker.js",
+  "FileApi/blob/Blob-slice.any.js",
+  "FileApi/blob/Blob-slice-overflow.any.js",
+  "FileApi/blob/Blob-stream.any.js",
+  "FileApi/blob/Blob-stream-byob-crash.html",
+  "FileApi/blob/Blob-stream-sync-xhr-crash.html",
```
guybedford, line 176: We don't support running HTML tests:

```suggestion
  "FileApi/blob/Blob-in-worker.worker.js",
  "FileApi/blob/Blob-slice.any.js",
  "FileApi/blob/Blob-slice-overflow.any.js",
  "FileApi/blob/Blob-stream.any.js",
```

guybedford (commented):
To update the WPT coverage, you need to create a JSON file in https://github.com/bytecodealliance/StarlingMonkey/tree/main/tests/wpt-harness/expectations for the corresponding tests you add, corresponding to the same test name but as `.any.js.json` instead of `.any.js`.

These files can be auto-generated with the `--update-expectations` WPT flag in https://github.com/bytecodealliance/StarlingMonkey/blob/main/tests/wpt-harness/run-wpt.mjs#L79.

WPT flags can be set when doing a manual WPT run via `WPT_FLAGS="--update-expectations"` which is covered at the end of https://github.com/bytecodealliance/StarlingMonkey?tab=readme-ov-file#web-platform-tests.

tests/wpt-harness/tests.json
andreiltd: Thanks. It turned out the file path is `FileAPI` not `FileApi` 🤦 I fixed it with this:

https://github.com/bytecodealliance/StarlingMonkey/pull/169/commits/701606692384fd60f4f4863e6b751bd604eca94e

builtins/web/blob.cpp
```diff
@@ -0,0 +1,499 @@
...
+  if (!promise) {
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
+
+  args.rval().setObject(*promise);
+
+  auto blob = Blob::blob(self);
+  auto chars = JS::UTF8Chars((char *)blob->data(), blob->size());
+
+  JS::RootedString str(cx, JS_NewStringCopyUTF8N(cx, chars));
```
andreiltd, line 241: The return value should be a result of decoding UTF8 on the blob as per https://w3c.github.io/FileAPI/#dom-blob-text. It looks like decoding function is fallible which makes some of the tests to fail. 

For example this test: https://github.com/web-platform-tests/wpt/blob/master/FileAPI/blob/Blob-text.any.js#L45

andreiltd:
Is `ReadableStream` something we support already or does it depend on: https://github.com/bytecodealliance/StarlingMonkey/issues/126 ?

tschneidereit:
> Is `ReadableStream` something we support already or does it depend on: #126 ?

It absolutely is, yes. We have some gaps on byte streams specifically, but `Blob#stream()` should be fully supportable.

You probably need to make use of [NativeStreamSource](https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/streams/native-stream-source.h#L9), which we're currently using for `body` streams.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,488 @@
...
+  if (is_typed_array) {
+    // append typed array value directly...
+    return append_value(cx, self, value);
+   } else if (is_iterable) {
+    // if the object is an iterable, walk over its elements...
+    JS::Rooted<JS::Value> item(cx);
+    while (true) {
+      bool done;
+
+      if (!it.next(&item, &done)) {
```
andreiltd, line 352: The iterator fails here for code like this:

```js
  var obj = {
    [Symbol.iterator]: Array.prototype[Symbol.iterator],
    get length() { throw test_error; }
  };
  assert_throws_exactly(test_error, function() {
    new Blob(obj);
  });
```
I'm not sure if this require a different approach or perhaps fixing in SpiderMonkey.

andreiltd: Currently, the Blob constructor in this code throws a `TypeError`, but the test expects the length `getter` to be invoked, which should throw `test_error`:

```
Threw object "TypeError: Blob.constructor: blobParts must be an object" but expected it to throw object "[object Object]"
```

The issue is that the length function is never called because the iterator fails on the `it.next()` function.

tschneidereit: Ah, I see. I'll make some suggested changes to the code that I think should resolve this, though some tweaks might be required.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+      if (!append_value(cx, self, item)) {
+        return false;
+      }
+    }
+
+    return true;
+  } else {
+    // non-objects are not allowed for the blobParts
+    return false;
```
tschneidereit, line 367:
```suggestion
    return api::throw_error(cx, api::Errors::TypeError, "Blob.constructor", "blobParts", "be an object");
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,488 @@
...
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), PrivateValue(JS_GetEmptyString(cx)));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Endings), JS::Int32Value(Blob::Endings::Transparent));
+
+  // walk the blob parts and write it to concatenated buffer
+  if (blobParts.isNull()) {
+    return api::throw_error(cx, api::Errors::TypeError, "Blob.constructor", "blobParts", "be an object");
+  }
+
+  if (!blobParts.isUndefined() && !init_blob_parts(cx, self, blobParts)) {
+    return api::throw_error(cx, api::Errors::TypeError, "Blob.constructor", "blobParts", "be an object");
```
tschneidereit, line 462:
```suggestion
    return false;
```

builtins/web/blob.cpp
tschneidereit: I think those two suggestions should make this work. Essentially, ensure that we only throw the "must be an object" error when that is actually the issue, and otherwise keep whatever pending exception we have active.

andreiltd: Oh nice, yeah this fixed few failing tests, thanks!

guybedford:
Note that #126 is not a blocker here - rather #126 is tracking additional readable stream features to the ones we already support.

andreiltd:
> Note that #126 is not a blocker here - rather #126 is tracking additional readable stream features to the ones we already support.

Right, I saw #126 and jumped to a conclusion. I’ll add the streaming API :)

andreiltd:
> > Is `ReadableStream` something we support already or does it depend on: #126 ?
> 
> It absolutely is, yes. We have some gaps on byte streams specifically, but `Blob#stream()` should be fully supportable.
> 
> You probably need to make use of [NativeStreamSource](https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/streams/native-stream-source.h#L9), which we're currently using for `body` streams.

Hey @tschneidereit I looked at the body stream implementation, but I'm not sure how to adapt it for a `Blob-readable` stream. AFAICT, the `NativeStreamSource` requires a pulling function. In the existing code, this function initiates an async task to pull data. I assume I would need something similar. The part I'm uncertain about is that the `body` async task uses a `pollable` [handle](https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/fetch/request-response.cpp#L70) to drive the future execution. What would be the equivalent in the case of a `Blob`?

guybedford:
@andreiltd you can follow the same process as `request-response.cpp` here in defining a pull algorithm function, but then the algorithm itself can just copy the buffers synchronously down to backpressure. None of these types touch the `AsyncTask` model system, the only promise comes from the potential backpressure which should be fine to handle in exactly the same way as for request. That is, there is no `BodyFutureTask`, just a `body_source_pull_algorithm`, and it should be a fairly degenerate case of this given there is no waiting needed on the incoming buffer.

andreiltd:
Thanks @guybedford . I considered using a task because this is what spec mentions as well: https://w3c.github.io/FileAPI/#blob-get-stream but I probably just misunderstood what needs to be done here :) Thanks for clarification!

guybedford:
Ahh I see, right we have no concept of immediate tasks to add to the task list which do not depend on WASI polls currently.

One option might be to use `EnqueueJob` to interleave with the Spidermonkey naive task system.

Alternatively we do have the concept of tasks having a `deadline()`. It might make sense to support a new `ImmediateTask` subclass of `AsyncTask` which treats it's handle as always invalid and its deadline as the current time, and then to update the WASI task `::select` implementation to resolve such immediate tasks before the main task queue.

Either sounds to me like it would get us semantic correctness in draining the microtasks first before draining tasks, and Fastly's `::select` implementation already does this exactly as well.

Perhaps @tschneidereit may have more thoughts on this as well.

andreiltd:
Hi @guybedford I tried to hack basic stream support in [this commit](https://github.com/bytecodealliance/StarlingMonkey/pull/169/commits/ce0899198ff09632ffe37c6f4fe1eaabc36047c9), just to get something working. 

It looks like the `event_loop` is not running. Would my code be also responsible for running the event loop? If so how do I run it without blocking?

guybedford:
That looks like a good start to me. Perhaps `IMMEDIATE_TASK_HANDLE` is a better term?

For the event loop - you can use `event.waitUntil(promise)` to make the event loop extend until that promise has resolved, and then have the stream work done in an async function that provides that promise.

andreiltd:
Hey @guybedford So I have an issue where the `stream` platform tests hang while waiting for the next chunk from the stream:

https://github.com/web-platform-tests/wpt/blob/master/FileAPI/blob/Blob-stream.any.js#L34

The event loop only iterates once. This is somewhat expected, as the task is always removed from the queue after it runs, which prevents us from closing the stream and returning `done`:

https://github.com/bytecodealliance/StarlingMonkey/blob/main/runtime/event_loop.cpp#L90

I was assuming that the task can be stateful and its entry function will be repeatedly called until it is cancelled? This seems not to be the case :)

guybedford:
Hey @andreiltd, event loop tasks go on the queue, then get immediately removed from the queue when invoked. So they only run once. The reason for the stream task here is one of timing - to queue work to happen as part of the task event loop rather than to be immediately synchronously blocking.

The way the event loop runs in StarlingMonkey is by continuing to drive the event loop until the top-level promise for the request/response handler has resolved. The only way to extend the event loop is to explicitly extend this promise or to extend with `event.waitUntil`.

The `promise_test` handler for the WPT tests should be ensuring that the promise_test promise delays the main event loop on the test promise. If that is not the case it could be a bug in the runner wrapper, but I'd be surprised if that is the case.

If you're still stuck, ping me and let me know if you want to pair on this at some point.

andreiltd:
Hey @guybedford . Sorry for late reply, I got distracted with some other work. 

Thanks for the explanation! The issue was that I assumed my task instance would be scheduled repeatedly, so I kept the reading state within the task instance. As a result, I kept starting the reading process from the beginning and never finished.

I moved the reader state to `Blob` class, and everything started working. 🎉

guybedford:
@andreiltd great to hear it's coming together. Yes you would need to move the reader state off the blob itself I think. It should be possible to augment the task with the reader value then just ensure it is always passed as context through the various promises after the task is picked up. The wiring will be specific to the code paths.

If we get that sorted will this PR be just about there?

andreiltd:
> If we get that sorted will this PR be just about there?

I think so, yes.

tests/wpt-harness/expectations/FileAPI/blob/Blob-constructor.any.js.json
```diff
@@ -0,0 +1,191 @@
+{
+  "Blob interface object": {
+    "status": "FAIL"
```
andreiltd, line 3: I'm not sure where should interface object be defined. This test fails on asserting the `Blob.length` which should be 0, but actual value is 1.

```
  assert_true("Blob" in globalThis, "globalThis should have a Blob property.");
  assert_equals(Blob.length, 0, "Blob.length should be 0.");
  assert_true(Blob instanceof Function, "Blob should be a function.");

```

builtins/web/blob.h
```diff
@@ -0,0 +1,87 @@
...
+  static bool size_get(JSContext *cx, unsigned argc, JS::Value *vp);
+  static bool type_get(JSContext *cx, unsigned argc, JS::Value *vp);
+
+public:
+  static constexpr const char *class_name = "Blob";
+
+  static const JSFunctionSpec static_methods[];
+  static const JSPropertySpec static_properties[];
+  static const JSFunctionSpec methods[];
+  static const JSPropertySpec properties[];
```
guybedford, line 50: I believe something like this should get the constructor length to update:

```
static constexpr int ctor_length = 0;
```

tests/wpt-harness/expectations/FileAPI/blob/Blob-stream.any.js.json
```diff
@@ -0,0 +1,20 @@
...
+    "status": "PASS"
+  },
+  "Blob.stream() garbage collection of blob shouldn't break stream consumption": {
+    "status": "PASS"
+  },
+  "Blob.stream() garbage collection of stream shouldn't break stream consumption": {
+    "status": "PASS"
+  },
+  "Reading Blob.stream() with BYOB reader": {
+    "status": "FAIL"
```
guybedford, line 18: Out of interest, what is the issue here?

andreiltd:
Something that isn't obvious to me when I look at the `Stream.h` API is how to set up the stream with byte reading support. 

I get this error when I try to get `BYOB` reader:

```
const stream = blob.stream();
const reader = stream.getReader({ mode: "byob" });

"TypeError: can't get a BYOB reader for a non-byte stream"
```

guybedford:
Is this ready for final review then?

tests/wpt-harness/expectations/FileAPI/blob/Blob-stream.any.js.json
andreiltd: Oh, bad timing on my part. I just asked the question regarding this failure :)

guybedford: I'd have to refresh my BYOB knowledge, but it seems like that error is when this check fails:

```
if (!unwrappedStream->controller()->is<js::ReadableByteStreamController>()) {
```

That is, I think the issue is that we don't support byob streams for native stream sources?

guybedford: Added issue https://github.com/bytecodealliance/StarlingMonkey/issues/176.

andreiltd:
> Is this ready for final review then?

I think it would be good to fix two failing tests:
- A Uint8Array object should be treated as a sequence for the blobParts argument,
- Reading Blob.stream() with BYOB reader

But I'm happy to work on those after we merge this one if that's OK for you. I don't expect the fixes will change the shape of the code drastically.

guybedford:
> A Uint8Array object should be treated as a sequence for the blobParts argument,

Is there much to do here still for this one?

andreiltd:
> > A Uint8Array object should be treated as a sequence for the blobParts argument,
> 
> Is there much to do here still for this one?

I think there is something wrong with text encoding for this particular input, see the code below:

```
const blob = new Blob(new Uint8Array([1, 2, 3]));
const text = await blob.text();
assert_equals(text, "123");

1:   NAME:    A Uint8Array object should be treated as a sequence for the blobParts argument.
1:   MESSAGE: assert_equals: expected "123" but got "\x01\x02\x03"
```
This is a bit confusing because all the other 'text' tests are passing, and they often use `Uint8Array` as input.

andreiltd:
> > > A Uint8Array object should be treated as a sequence for the blobParts argument,
> > 
> > 
> > Is there much to do here still for this one?
> 
> I think there is something wrong with text encoding for this particular input, see the code below:
> 
> ```js
> const blob = new Blob(new Uint8Array([1, 2, 3]));
> const text = await blob.text();
> assert_equals(text, "123");
> ```
> 
> ```
> 1:   NAME:    A Uint8Array object should be treated as a sequence for the blobParts argument.
> 1:   MESSAGE: assert_equals: expected "123" but got "\x01\x02\x03"
> ```
> 
> This is a bit confusing because all the other 'text' tests are passing, and they often use `Uint8Array` as input.

OK, this is fixed now. It’s a bit unintuitive, but the way it should work is that if you pass a typed array directly to the constructor, each element should be converted to a string and appended to the blob data, rather than treating the array as binary data.

If you want to treat a typed array as binary data, it needs to be wrapped as an element of an array. For example:

```js
const arr = new Uint8Array([1, 2, 3]);
const blob1 = new Blob(arr);    // This will iterate over the elements of [1, 2, 3] and convert each element to a string.
const blob2 = new Blob([arr]);  // This will iterate over [arr] and append the raw bytes from arr.
```

builtins/web/blob.h
```diff
@@ -0,0 +1,88 @@
...
+#include "extension-api.h"
+#include "js/AllocPolicy.h"
+#include "js/TypeDecls.h"
+
+namespace builtins {
+namespace web {
+namespace blob {
+
+class BlobReader {
+  std::span<const uint8_t> data_;
```
guybedford, line 14: Trying to understand the memory model here.

The blob holds onto its readers, and the stream task references the blob, so readers are guaranteed to be available while streaming, so long as there is an active stream task.

When the stream task is drained, I assume there is JS code that runs and has access to the buffer from the reader.

In this state, when there is no stream task on the task queue, and the user is directly accessing the blob buffer, what keeps the blob alive then?

guybedford (commented):
Just one question about the memory model here, but otherwise this looks great to me.

builtins/web/blob.h
guybedford: That is, I'd have expected we'd need a strong GC reference from the stream backing store to the blob?

andreiltd: That's a great question. The `BlobReader` is only ever referenced in the context of `StreamTask`. What I believe keeps the blob alive is the fact that `StreamTask` maintains a reference to the blob like this:

```
    Heap<JSObject *> source_;  // -> reference to blob
```

`StreamTask` manages readers by either adding a reader state for a given stream if it doesn’t exist yet, updating or removing the reader when reading is finished. Essentially, the blob only provides a persistent reader context for the short-lived tasks.

That being said, I believe `BlobReader` and `readers` method should be removed from the public interface, as it has no meaning outside of `StreamTask` and should instead be an implementation detail of how tasks work. Otherwise, as you noticed, user could hold on to the `ReaderState` that has been invalidated.

guybedford: My concern was more the converse - if I read one chunk of a blob stream, but don't read the next one, and I'm holding a handle to the stream reader but let the blob itself be GC'd, then there's no "task" being waited on so there's no StreamTask in play - so there's nothing holding onto the blob and underlying data store, but the stream is still being held.

guybedford: In other words, the stream handler needs to hold a strong reference to the blob's data store or blob itself I think?

andreiltd: Right, I kind of assumed that `NativeStream` will prevent garbage collection of a blob while it exists but I can double check that :) There are also tests for this specific scenario, both passing:

```
  "Blob.stream() garbage collection of blob shouldn't break stream consumption"
  "Blob.stream() garbage collection of stream shouldn't break stream consumption"
```

guybedford: Right, I'm just asking for clarity on the model - where is the reference from `NativeStream` to blob that ensures this?

We don't currently have proper GC test support so it's tricky to verify these behaviours. In due course we will though, so it would suck if when we start turning on agressive GC for testing a whole bunch of subsystems start breaking.

andreiltd: In the `Blob::stream()` method we create a `NativeStreamSource` like this:

```cpp
  auto native_stream = streams::NativeStreamSource::create(
    cx, self, JS::UndefinedHandleValue, stream_pull, stream_cancel);
```

where `self` is a blob `JS::HandleObject`. Then `NativeStreamSource` keeps the handle as a stream owner in the slot:

```cpp
JSObject *NativeStreamSource::create(JSContext *cx, JS::HandleObject owner,
                                     JS::HandleValue startPromise,
                                     PullAlgorithmImplementation *pull,
                                     CancelAlgorithmImplementation *cancel) {
  JS::RootedObject source(cx, JS_NewObjectWithGivenProto(cx, &class_, proto_obj));
  if (!source)
    return nullptr;

  JS::SetReservedSlot(source, Slots::Owner, JS::ObjectValue(*owner));
  ...
}
```

Will this ensure that the lifetime of the blob is at least as long as that of the stream?

guybedford: Ahh okay that sounds like it resolves my question then. In addition the buffers from the stream read are by copy so there's no references to worry about there either. Thanks for talking this through.

guybedford (approved):
Will let @tschneidereit review before landing, but this is amazing work, thank you!

andreiltd:
Thanks @guybedford. I guess the follow up is to support `Request#blob` so we can close https://github.com/bytecodealliance/StarlingMonkey/issues/121?

guybedford:
Yeah that would be a fantastic follow-up.

host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -5,9 +5,18 @@
...
+
+  for (size_t idx = 0; idx < count; ++idx) {
+    auto *task = tasks.at(idx);
+    auto id = task->id();
+
+    if (id == IMMEDIATE_TASK_HANDLE) {
+      return idx;
+    } else {
+      handles.emplace_back(id);
+    }
```
tschneidereit, line 17: While this is the obvious way of supporting internal tasks, unfortunately this way of integrating internal tasks means that they will always be run as much as possible, starving other tasks that might also be ready. E.g. if you're processing a large `Blob` while concurrently doing a `fetch`, the response to the `fetch` will only be processed once the entire `Blob` has been processed, instead of whenever it's actually available.

I think there's a decent way of handling this, though: if we encounter an `IMMEDIATE_TASK_HANDLE` in this loop, we can check if we have any other entries in `handles` already. If so, we can add an additional pollable that'll always be ready immediately, and do a poll of the entries before the `IMMEDIATE_TASK_HANDLE`. If any of those are ready, return the first one, otherwise return the `IMMEDIATE_TASK_HANDLE`.

I haven't tested this, but I think if you replace the entire `api::AsyncTask::select` method with the following code, things should either work, or be very close to working:

```cpp
static optional<wasi_clocks_monotonic_clock_own_pollable_t> immediately_ready;

size_t poll_handles(vector<WASIHandle<host_api::Pollable>::Borrowed> handles) {
  auto list = list_borrow_pollable_t{handles.data(), handles.size()};
  bindings_list_u32_t result{nullptr, 0};
  wasi_io_poll_poll(&list, &result);
  MOZ_ASSERT(result.len > 0);
  const auto ready_index = result.ptr[0];
  free(result.ptr);
  return ready_index;
}

size_t api::AsyncTask::select(std::vector<AsyncTask *> &tasks) {
  auto count = tasks.size();
  vector<WASIHandle<host_api::Pollable>::Borrowed> handles;
  for (size_t idx = 0; idx < count; ++idx) {
    auto *task = tasks.at(idx);
    auto id = task->id();

    if (id == IMMEDIATE_TASK_HANDLE) {
      if (handles.size() > 0) {
        if (!immediately_ready) {
          immediately_ready = wasi_clocks_monotonic_clock_subscribe_duration(0);
        }
        handles.emplace_back(immediately_ready.value().__handle);
        size_t ready_index = poll_handles(std::move(handles));
        if (ready_index <= idx) {
          return ready_index;
        }
      }
      return idx;
    }
    handles.emplace_back(id);
  }

  return poll_handles(std::move(handles));
}
```

tschneidereit (commented):
Hey Tomasz, my apologies for the delay here; I've been swamped with other stuff.

I still haven't been able to give this a full review, but just left an in-line comment on one key bit around event loop integration. It'd be great if you could look at that, and see if my suggestion works. If it does, adding a test for processing `Blob` events concurrently with other tasks would be fantastic, too.

andreiltd:
> Hey Tomasz, my apologies for the delay here; I've been swamped with other stuff.

No worries! I understand being swamped all too well :) Thanks for suggestions, I will give it a try!

andreiltd:
BTW, regarding tasks, we have a bunch of code that adds dynamically created tasks to the queue like this:

```cpp
ENGINE->queue_async_task(new BodyFutureTask(source));
```

I couldn't find the code where we actually release the task memory. Is it somehow GC manged? It doesn't look like it is?

andreiltd:
The suggested code seems to work. Thanks! But it looks like there is something wrong with the streaming for large blobs (>10M). I've added the test that creates a blob and read it in chunks. There is also an interval task running concurrently. This all seems to be working excecpt if I increase the size of the blob to 10M or 100M I get the size mismatch between the amount of bytes read from the stream and actual blob size. The stream size is always larger. I will have to debug this :)

andreiltd:
> But it looks like there is something wrong with the streaming for large blobs (>10M). 
I think I’ve got to the root of the problem: I’m storing references to GC-managed objects in the blob slot. This slot is a map that allows lookup of the reader state using the stream pointer.:

```
  using ReadersMap = js::HashMap<JSObject*, BlobReader, js::PointerHasher<JSObject*>, js::MallocAllocPolicy>;
```

This long running test triggers a GC run which invalidates the stored pointers and as a result it restarts the reading of the stream. I guess if I want to fix that I need to change the references to `Heap<JSObject *>` and implement `trace` on the `Blob`.  This mean I probably need to define something like `TraceableBuiltinImpl` so that I can connect `trace` method to the class ops.

host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -2,29 +2,43 @@
...
     if (id == IMMEDIATE_TASK_HANDLE) {
+      if (handles.size() > 0) {
+        if (!immediately_ready) {
+          immediately_ready = wasi_clocks_monotonic_clock_subscribe_duration(0);
+        }
+        handles.emplace_back(immediately_ready.value().__handle);
+        size_t ready_index = poll_handles(std::move(handles));
+        if (ready_index <= idx) {
+          return ready_index;
+        }
```
tschneidereit, line 34: I just realized that this is subtly wrong, my apologies. Instead of comparing to `idx` here, we need to check if `ready_index` is lower than `handles.size() - 1`. I.e., whether the first handle that's ready is any other than `immediately_ready`. And for that to work, we need to store `handles.size()` before the call to `poll_handles`, because we lose ownership of the vector at that point.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+// https://w3c.github.io/FileAPI/#convert-line-endings-to-native
+std::string convert_line_endings_to_native(std::string_view s) {
+  std::string native_line_ending = "\n";
+#ifdef _WIN32
+  native_line_ending = "\r\n";
+#endif
+
+  std::string result;
+  std::regex re(R"(\r\n|\r|\n)");
+  std::regex_replace(std::back_inserter(result), s.begin(), s.end(), re, native_line_ending);
```
tschneidereit, line 118: I'd prefer not pulling in `std::regex` for this, because it's pretty heavy. Could you instead do this as a manual loop? Basically, find each instance of `\r`, check if the next char is `\n`, then either replace or remove `\r` accordingly?

(This of course only works for non-Windows. Currently, that doesn't matter, since we only compile for Wasm. If you want to prepare things for a possible future in which we have native builds, you could add an `#ifdef` implementing the logic for that—up to you.)

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+public:
+  explicit StreamTask(const HandleObject source, api::Engine *engine)  : source_(source) {
+    engine->incr_event_loop_interest();
+    handle_ = IMMEDIATE_TASK_HANDLE;
+  }
+
+  [[nodiscard]] bool run(api::Engine *engine) override {
+    JSContext *cx = engine->cx();
+    RootedObject owner(cx, streams::NativeStreamSource::owner(source_));
+    RootedObject controller(cx, streams::NativeStreamSource::controller(source_));
```
tschneidereit, line 167: Instead of getting the controller here and then using `JS::Call` for the `close` and `enqueue` operations below, you can use `JS::ReadableStreamClose` and `JS::ReadableStreamEnqueue` on `source_` directly, I think.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+    return cancel(engine);
+  }
+
+  [[nodiscard]] bool cancel(api::Engine *engine) override {
+    handle_ = INVALID_POLLABLE_HANDLE;
+    engine->decr_event_loop_interest();
+    return true;
+  }
+
+  void trace(JSTracer *trc) override { TraceEdge(trc, &source_, "source for future"); }
```
tschneidereit, line 214:
```suggestion
  void trace(JSTracer *trc) override { TraceEdge(trc, &source_, "Source for Blob StreamTask"); }
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+  return true;
+}
+
+bool Blob::bytes(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  JS::RootedObject promise(cx, JS::NewPromiseObject(cx, nullptr));
+  if (!promise) {
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
tschneidereit, line 272:
```suggestion
  if (!promise) {
    return false;
  }
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+  return array_buffer;
+}
+
+bool Blob::arrayBuffer(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  JS::RootedObject promise(cx, JS::NewPromiseObject(cx, nullptr));
+  if (!promise) {
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
tschneidereit, line 250: If we weren't able to create a `Promise` object here, we'll pretty certainly also not be able to create on in `ReturnPromiseRejectedWithPendingError`. Given that, I think it makes sense to just return `false` here.
```suggestion
  if (!promise) {
    return false;
  }
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+
+  if (args.hasDefined(2)) {
+    HandleValue contentType_val = args.get(2);
+    if (!(contentType = normalize_type(cx, contentType_val))) {
+      return false;
+    }
+  }
+
+  // A negative value, is treated as an offset from the end of the Blob toward the beginning
```
tschneidereit, line 325:
```suggestion
  // A negative value is treated as an offset from the end of the Blob toward the beginning.
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+  return true;
+}
+
+bool Blob::text(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0)
+
+  JS::RootedObject promise(cx, JS::NewPromiseObject(cx, nullptr));
+  if (!promise) {
+    return ReturnPromiseRejectedWithPendingError(cx, args);
+  }
```
tschneidereit, line 376:
```suggestion
  if (!promise) {
    return false;
  }
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+    JS::RootedValue type(cx);
+    if (!JS_GetProperty(cx, opts, "type", &type)) {
+      return false;
+    }
+
+    auto type_str = normalize_type(cx, type);
+    if (!type_str) {
+      return false;
+    }
+    SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), PrivateValue(type_str));
```
tschneidereit, line 623: You should store `type_str` as a string instead of a private value to ensure that the GC knows about it. Changing `PrivateValue` to `StringValue` here and adapting the other sites accessing `Slots::Type` should be all that's needed.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+  RootedValue blobParts(cx, args.get(0));
+  RootedValue opts(cx, args.get(1));
+  RootedObject self(cx, JS_NewObjectForConstructor(cx, &class_, args));
+
+  if (!self) {
+    return false;
+  }
+
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Data), PrivateValue(new std::vector<uint8_t>()));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), PrivateValue(JS_GetEmptyString(cx)));
```
tschneidereit, line 654:
```suggestion
  SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), JS_GetEmptyStringValue(cx));
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+    if (!type_str) {
+      return false;
+    }
+    SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), PrivateValue(type_str));
+  }
+
+  return true;
+}
+
+JSObject *Blob::create(JSContext *cx, std::unique_ptr<Blob::ByteBuffer> data, JSString *type) {
```
tschneidereit, line 629: This should take a `HandleString` instead of a raw `JSString*` to ensure proper rooting. Please check all other functions to ensure that you don't pass raw pointers to GC objects like this.

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+  if (!self) {
+    return false;
+  }
+
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Data), PrivateValue(new std::vector<uint8_t>()));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Type), PrivateValue(JS_GetEmptyString(cx)));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Endings), JS::Int32Value(LineEndings::Transparent));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Readers), PrivateValue(new ReadersMap));
+
+  // walk the blob parts and write it to concatenated buffer
```
tschneidereit, line 658:
```suggestion
  // Walk the blob parts and append them to the blob's buffer.
```

builtins/web/blob.cpp
```diff
@@ -0,0 +1,703 @@
...
+
+void Blob::finalize(JS::GCContext *gcx, JSObject *self) {
+  auto blob = Blob::blob(self);
+  if (blob) {
+    free(blob);
+  }
+
+  auto readers = Blob::readers(self);
+  if (readers) {
+    free(readers);
```
tschneidereit, line 687: IIUC, a `BlobReader` doesn't own the `span` it holds, but instead is a view onto the `Blob`'s underlying buffer, right? If so, this seems correct, but would be good to document as part of the `BlobReader` definition.

builtins/web/blob.h
tschneidereit: Thank you for the discussion here! It would be great if we could have the key properties of the memory model documented in code somewhere, as this wasn't clear to me at all, either. (See my question in `Blob::finalize` above.)

tschneidereit (changes requested):
This is really great, thank you so much for the excellent work here!

I left a bunch of comments, most of which should be addressed, but it's very close!

builtins/web/blob.cpp
andreiltd: The blob buffer is something I plan to change, but in a follow-up PR. I already have a sketch for this in the Request#blob API work. Specifically, I’d like to switch from `std::vector` to `mozilla::Vector`. The reason is that mozilla::Vector provides optimizations that aren’t possible with `std::vector`, such as building a vector from raw parts, which allows creating a blob without copying buffers but simply by transferring the ownership.

builtins/web/blob.cpp
andreiltd: I wasn't able to make this working. If I try to use Close/Enque directly this fails on some downcasting assertions. I think in order for this to work I need to have a reference to a `DefaultReadableStream` object, but I wouldn't know how to get this from `source_`.

builtins/web/blob.h
andreiltd: Would it be OK if I document the model as a part of PR that switches from `std::vector` to mozilla::Vector` as this will probably change the reader quiet a bit. See my comment here: https://github.com/bytecodealliance/StarlingMonkey/pull/169#discussion_r1862493507

builtins/web/blob.cpp
tschneidereit: Oh I see; you're not storing the stream itself anywhere, so you don't have access to it here. And `NativeStreamSource::stream` isn't any help, because it assumes that the owner will be a `BodySource` :(

It might be easiest to fix that by storing the stream itself on the source explicitly. If you're up for it, you could do that, and instead remove the `Controller` slot, because that can be retrieved from the stream, but not the other way around.

The reason all of this is important is that otherwise content can change what's happening here by setting different functions for `ReadableStreamDefaultController.prototype.{close, enqueue}`, which we absolutely don't want.

builtins/web/blob.h
tschneidereit: Yes, absolutely!

builtins/web/blob.cpp
andreiltd: Right, the initialization order now is as follows:

```cpp
  JS::RootedObject native_source(cx, streams::NativeStreamSource::create(
    cx, self, JS::UndefinedHandleValue, stream_pull, stream_cancel));

  JS::RootedObject source_stream(cx, JS::NewReadableDefaultStreamObject(cx, native_source, nullptr, 0.0));
```

So what you're proposing is to move the initialization of the default stream to `NativeStreamSource::create` and store it in a slot there?

tschneidereit: I think that'd make sense, yes

tschneidereit: I feared that might be the case :( What I'd do from here is to either reduce the test case / add `console.log` statements to it to get more information, or to add debugging output using the `DBG` macro to the C++ code.

But yes, that all should be for next week. For now, I hope you have a great weekend!

andreiltd: OK, I think I've figured this out. The issue was that I initialized `native-stream` slots after passing the instance handle to `NewReadableDefaultStream` constructor.

builtins/web/blob.cpp
```diff
@@ -677,19 +696,24 @@ bool Blob::init_class(JSContext *cx, JS::HandleObject global) {
 }
 
 void Blob::finalize(JS::GCContext *gcx, JSObject *self) {
-  auto blob = Blob::blob(self);
+  auto blob = static_cast<std::vector<uint8_t> *>(
```
tschneidereit, line 699:
```suggestion
  auto blob = static_cast<std::vector<uint8_t> *>(
```
```suggestion

  MOZ_ASSERT(is_instance(self));
  auto blob = static_cast<std::vector<uint8_t> *>(
```

builtins/web/blob.cpp
```diff
@@ -677,19 +696,24 @@ bool Blob::init_class(JSContext *cx, JS::HandleObject global) {
...
+  auto readers = static_cast<ReadersMap *>(
+      JS::GetReservedSlot(self, static_cast<size_t>(Blob::Slots::Readers)).toPrivate());
   if (readers) {
     free(readers);
   }
 }
 
 void Blob::trace(JSTracer* trc, JSObject *self) {
-  auto readers = Blob::readers(self);
+  auto readers = static_cast<ReadersMap *>(
```
tschneidereit, line 714:
```suggestion
  auto readers = static_cast<ReadersMap *>(
```
```suggestion
  MOZ_ASSERT(is_instance(self));
  auto readers = static_cast<ReadersMap *>(
```

builtins/web/streams/native-stream-source.h
```diff
@@ -51,7 +51,8 @@ class NativeStreamSource : public BuiltinNoConstructor<NativeStreamSource> {
   static bool pull(JSContext *cx, unsigned argc, JS::Value *vp);
   static bool cancel(JSContext *cx, unsigned argc, JS::Value *vp);
   static JSObject *create(JSContext *cx, JS::HandleObject owner, JS::HandleValue startPromise,
-                          PullAlgorithmImplementation *pull, CancelAlgorithmImplementation *cancel);
+                          PullAlgorithmImplementation *pull, CancelAlgorithmImplementation *cancel,
+                          JS::HandleFunction size = nullptr, double highWaterMark = 0.0);
```
tschneidereit, line 55: The comment in `RequestOrResponse::create_body_stream` was confusing to me until I realized that not passing a HWM uses an `0.0` default. I think in this case it'd be clearer to make that explicit and not have a default value here at all.
```suggestion
                          JS::HandleFunction size = nullptr, double highWaterMark = 0.0);
```
```suggestion
                          JS::HandleFunction size = nullptr, double highWaterMark);
```

builtins/web/streams/native-stream-source.h
```diff
@@ -35,10 +35,10 @@ class NativeStreamSource : public BuiltinNoConstructor<NativeStreamSource> {
                                              JS::HandleValue reason);
   static JSObject *owner(JSObject *self);
   static JSObject *stream(JSObject *self);
+  static JSObject *default_stream(JSObject *self);
```
tschneidereit, line 38: Do we still need both of these, or could we replace the logic inside of `stream` with that of `default_stream` instead of introducing a second function?

If we do still need both, then I think we should either very carefully document the difference between the two, or change things enough so we stop needing both. Otherwise this seems like a recipe for bugs later on: I don't think I would have any idea of when to use one over the other, personally.

tschneidereit (changes requested):
Great job tracking down that issue!

This is very, very close, with just a few more changes needed. Some of those are because I gave unclear feedback earlier—my apologies for that!

builtins/web/streams/native-stream-source.h
andreiltd: I would think `0.0`  is still a sensible default considering all builtin code sets `HWM` to that value. Maybe we can move the comment to the `native-stream` implementation that would motivate why we choose to overwrite the default value from `1.0` to `0.0`.

I'm fine with either, although I would prefer having argument list in this function somewhat similar to `NewReadableStreamDefaultObject`. If we remove default value for `HWM` I guess we also need to change the order of `size` and `HWM` as the compiler won't let us having default argument followed by argument with explicit value.

builtins/web/streams/native-stream-source.h
tschneidereit: Could you maybe add an assert like this to `NativeStreamSource::stream` and see if it's hit when running the test suite?
```cpp
MOZ_ASSERT(web::fetch::RequestOrResponse::body_stream(owner(self)) == default_stream(self));`

If it's not hit, it should be fine to replace `NativeStreamSource::stream` instead of adding `default_stream`

File: builtins/web/streams/native-stream-source.h (hunk 26 above)
Comment (tschneidereit) [reply builtins/web/streams/native-stream-source.h:55]: ah, good point about keeping things similar to `NewReadableStreamDefaultObject`.

I think I agree that moving the comment to `native-stream` makes sense. You'd then have to adjust it to not talk about http bodies anymore, and instead focus on explaining why for native streams a different default is chosen. But that seems like a good change to make

**Comment by guybedford** [comment]
@andreiltd have all the review comments here been addressed now?

**Comment by andreiltd** [comment]
> @andreiltd have all the review comments here been addressed now?

I think so, yes.

File: builtins/web/streams/native-stream-source.h (hunk 26 above)
Comment (andreiltd) [reply builtins/web/streams/native-stream-source.h:62]: OK the comment is now in the `natvie-stream` header file.

File: builtins/web/blob.cpp [hunk 28]
```diff
@@ -474,35 +474,40 @@ bool Blob::stream_pull(JSContext *cx, JS::CallArgs args, JS::HandleObject source
...
 }
 
-size_t Blob::blob_size(HandleObject self) {
+size_t Blob::blob_size(JSObject *self) {
+  MOZ_ASSERT(is_instance(self));
   auto blob = static_cast<std::vector<uint8_t> *>(
       JS::GetReservedSlot(self, static_cast<size_t>(Blob::Slots::Data)).toPrivate());
 
   MOZ_ASSERT(blob);
   return blob->size();
```
tschneidereit, line 492: I just realized that you can reduce this function to just be this:
```suggestion
  return blob(self)->size();
```

tschneidereit (approved):
Fantastic work, thank you so much! I really appreciate the attention to detail, and the patience in working through the feedback <3

builtins/web/blob.cpp
andreiltd: Nice!

guybedford:
Congrats on a fantastic first PR here 🎉

PR #171, opened 2024-10-29
Fix bugs in handling multiple incoming bodies piped through a `TransFormStream` to an outgoing body

tschneidereit:
This was pretty gnarly to work through, but I think came out reasonably clean. The two key pieces are:
- delaying the first read from an incoming stream that's been piped to a `TransformStream` until the latter is actually read from, and hence doesn't have backpressure applied anymore. This part guarantees that when the read would happen, we have the full pipeline in place and can hand things off to the host API instead of handling anything in the JS `fetch` API implementation.
- properly integrating these two pieces, and only closing the incoming body's `ReadableStream` once the host API reports that it's been dealt with to completion.

This PR additionally contains some fixes to how writes to bodies are handled, and some cleanup. These are both in separate patches which can be reviewed separately, if that seems easier.

### DIFF
--- a/host-apis/wasi-0.2.0/host_api.cpp
+++ b/host-apis/wasi-0.2.0/host_api.cpp
@@ -638,13 +638,8 @@ public:
                           HandleObject callback_receiver)
       : incoming_body_(incoming_body), outgoing_body_(outgoing_body), cb_(completion_callback),
         cb_receiver_(callback_receiver), state_(State::BlockedOnBoth) {
-    auto res = incoming_body_->subscribe();
-    MOZ_ASSERT(!res.is_err());
-    incoming_pollable_ = res.unwrap();
-
-    res = outgoing_body_->subscribe();
-    MOZ_ASSERT(!res.is_err());
-    outgoing_pollable_ = res.unwrap();
+    incoming_pollable_ = incoming_body_->subscribe().unwrap();
+    outgoing_pollable_ = outgoing_body_->subscribe().unwrap();
   }
 
   [[nodiscard]] bool run(api::Engine *engine) override {
--- a/host-apis/wasi-0.2.0/host_api.cpp
+++ b/host-apis/wasi-0.2.0/host_api.cpp
@@ -574,31 +574,69 @@ void HttpOutgoingBody::write(const uint8_t *bytes, size_t len) {
   MOZ_RELEASE_ASSERT(write_to_outgoing_body(borrow, bytes, len));
 }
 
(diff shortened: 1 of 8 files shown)


host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -574,31 +574,69 @@ void HttpOutgoingBody::write(const uint8_t *bytes, size_t len) {
...
-    auto bytes_to_write = std::min(len, static_cast<size_t>(capacity));
-    if (!write_to_outgoing_body(borrow, bytes, len)) {
-      return Result<Void>::err(154);
+    uint64_t capacity = res.unwrap();
+    MOZ_ASSERT(capacity >= 0);
+    auto bytes_to_write = std::min(bytes_.len - offset_, static_cast<size_t>(capacity));
+    outgoing_body_->write(bytes_.ptr.get() + offset_, bytes_to_write);
+    offset_ += bytes_to_write;
+    if (offset_ < bytes_.len) {
+      engine->queue_async_task(this);
```
guybedford, line 606: Is it possible for the capacity to increase after a write operation such that we should check the capacity again here and possibly loop?

My main concern would be in hosts where a low capacity is reported, that we might end up with a lot of async slowdown when sync loops should really be preferred?

Perhaps we should instead make the signal for polling an explicit capacity zero return per the WASI HTTP descriptions here?

guybedford (commented):
Looks great thanks for working on this. It's just a shame Fastly can't share the implementation of these improvements.

I was a little surprised the WPT had no coverage here to change, and tested the most recent WPT against this, but it seems that's just the way it goes.

host-apis/wasi-0.2.0/host_api.cpp
tschneidereit: That's a great point, yes. I changed this to run in a loop until either all bytes have been written or no more capacity is signaled.

tschneidereit:
> Looks great thanks for working on this. It's just a shame Fastly can't share the implementation of these improvements.

Fastly doesn't need this, as C@E has a hostcall for doing this kind of append natively and much more efficiently, [which the JS runtime makes use of](https://github.com/fastly/js-compute-runtime/pull/62). I'm working on adding an equivalent thing to `wasi:http`.

> I was a little surprised the WPT had no coverage here to change, and tested the most recent WPT against this, but it seems that's just the way it goes.

Same: I would've expected this to be covered somewhere and was surprised it's not.

Issue #172, opened 2024-10-30
WASI 0.2.1, 0.2.2 bindings


guybedford:
We should at the very least support outputting to 0.2.2 bindings at this point, which would be useful in ComponentizeJS to coalesce the final binary as we're currently building binaries with multiple version imports.

Should we support all versions separately or can we go straight to 0.2.2?

PR #174, opened 2024-11-01
Update to WASI@0.2.2

guybedford:
This adds a host implementation for WASI@0.2.2, and sets it to the default.

Resolves https://github.com/bytecodealliance/StarlingMonkey/issues/172.


tschneidereit:
Incidentally, I created a patch making it easier to support multiple WASI versions just recently. I just opened #175 to land that separate from the work I did it for (which you can see [here](https://github.com/bytecodealliance/StarlingMonkey/tree/wasi-http-body-append)), and think it'd probably make sense to rebase this PR on top of it to reduce code duplication.

Also, it'd be nice to document what changed in terms of functionality, if anything.

guybedford:
@tschneidereit I understand you have a better architecture in mind here, but for what it is worth I've rebased this PR to main now.

tschneidereit:
@guybedford I pushed a new commit here that implements the deduplication I [suggested earlier](pullrequestreview-2468387110). As you can see, the diff to `main` becomes substantially smaller with this change applied.

guybedford (commented):
Thanks, LGTM!

Happy to get the similar 0.2.3 going when we do the upgrade across Jco and ComponentizeJS soon.

tschneidereit (approved):
Approving this, but would like an r+ from @guybedford as well, given my substantial additions to the PR itself.

tschneidereit:
Actually, I can't select Guy as a formal reviewer, but he gave an LGTM above.

PR #175, opened 2024-11-03
Reorganize WASI 0.2.0 bindings to facilitate reuse

tschneidereit:
This patch doesn't introduce any functional changes: it's purely a change to how the WASI 0.2.0 host API implementation is organized. This will be used in further patches introducing more refactorings in service of more easily building WASI 0.2.x bindings making use of new WASI features.

Note that there's one part here that breaks embeddings providing their own host API: the `host_api.cmake` file has moved into the API implementation directory to enable more flexible organization of the implementation.


guybedford (approved):
So new hosts would just be expected to include their own host_api.cmake file, but that file could just be potentially empty right?

tschneidereit:
> So new hosts would just be expected to include their own host_api.cmake file, but that file could just be potentially empty right?

At least for now, the `host_api.cmake` file would need to be duplicated, though it could in principle just include another host API implementation's file. I'm open to ideas for how to reduce duplication further, though in practice I think much of what's in there will have to be slightly tweaked.

PR #177, opened 2024-11-25
Add Request and Response blob methods

andreiltd:
This patch adds `blob` API to `Request` and `Response` methods.

The patch also refactors the internal blob storage, changing it from `std::vector` to `mozilla::Vector`. This optimization allows the `Blob` to be created without copying data by transferring buffer ownership, as is the case for both `Response` and `Request` objects.

Closes: https://github.com/bytecodealliance/StarlingMonkey/issues/121

### DIFF
--- a/builtins/web/fetch/request-response.cpp
+++ b/builtins/web/fetch/request-response.cpp
@@ -549,7 +549,6 @@ bool RequestOrResponse::parse_body(JSContext *cx, JS::HandleObject self, JS::Uni
       return RejectPromiseWithPendingError(cx, result_promise);
     }
 
-    // We can drop `buf` as the data has been now copied over to blob
     result.setObject(*blob);
   } else {
     JS::RootedString text(cx, JS_NewStringCopyUTF8N(cx, JS::UTF8Chars(buf.get(), len)));
--- a/builtins/web/fetch/request-response.cpp
+++ b/builtins/web/fetch/request-response.cpp
@@ -27,6 +27,8 @@
 #include "js/experimental/TypedData.h"
 #pragma clang diagnostic pop
 
+using builtins::web::blob::Blob;
+
 namespace builtins::web::streams {
 
 bool NativeStreamSource::stream_is_body(JSContext *cx, JS::HandleObject stream) {
@@ -290,6 +292,7 @@ bool RequestOrResponse::extract_body(JSContext *cx, JS::HandleObject self,
   // We currently support five types of body inputs:
   // - byte sequence
   // - buffer source
+  // - Blob
   // - USV strings
   // - URLSearchParams
   // - ReadableStream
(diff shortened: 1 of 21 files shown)


builtins/web/fetch/request-response.cpp
```diff
@@ -546,6 +548,17 @@ bool RequestOrResponse::parse_body(JSContext *cx, JS::HandleObject self, JS::Uni
     }
     static_cast<void>(buf.release());
     result.setObject(*array_buffer);
+  } else if constexpr (result_type == RequestOrResponse::BodyReadResult::Blob) {
+    auto bytes = reinterpret_cast<uint8_t *>(buf.get());
+    auto data = std::make_unique<std::vector<uint8_t>>(bytes, bytes + len);
```
tschneidereit, line 553: It's a bit unfortunate that this has to copy the bytes again :/ Maybe open an issue about this and reference it as part of a TODO here?

(Unfortunately the only real option to fix this would I think be to stop using a `vector` in `Blob`, which isn't great. And solidly makes this an "optimization problem for later"—hence the suggestion to open an issue for it instead of fixing it here and now.)

andreiltd: Oh, see my comment here:

https://github.com/bytecodealliance/StarlingMonkey/pull/169#discussion_r1862493507

tschneidereit: Perfect, thank you!

andreiltd: As of the latest [commit](https://github.com/bytecodealliance/StarlingMonkey/pull/177/commits/43f2f9120a01080197f84b1984e6fe0e75ad6f2d) the buffer is not copied anymore 🎉

guybedford:
It would also be nice to see in this PR for `return new Response(blob)` setting the content-type header from the blob type (see the Blob path in https://fetch.spec.whatwg.org/#concept-bodyinit-extract).

guybedford:
I've also posted https://github.com/bytecodealliance/StarlingMonkey/issues/182 as a further follow-up.

andreiltd:
> Update: I was able to implement these for Fastly here - https://github.com/fastly/js-compute-runtime/commit/9b59afa9cb5b9b3f818226f9ca8676d5b40cf95b

Nice! This looks to me like an equivalent of doing:

```cpp
RootedValue blob_stream(cx);
Call(cx, body_obj, "stream", HandleValueArray::empty(), &blob_stream);
```

Is there a reason we don't want to use `Call`? Is this indirection not performant enough? If that's the case I should probably add API that allows calling blob methods like `stream`, `bytes` etc directly. WDYT?

andreiltd:
I've also added support for `Blob` in structured-clone but perhaps this should be in another PR. Let me know if you want me to split this.

guybedford:
@andreiltd yeah in general I'd avoid running through the high-level call for performance - we very much need to always keep performance top of mind in development on StarlingMonkey, because by default we are running at a performance penalty compared to other JS platforms, performance is of critical performance for this codebase.

guybedford (approved):

andreiltd:
That makes sense. I will make sure to add interface for calling Blob functions directly in the File API PR.

Issue #155, opened 2024-10-11
File API


guybedford:
https://w3c.github.io/FileAPI/#dfn-file

Should probably come after #121

PR #181, opened 2024-12-04
Add File interface

andreiltd:
https://w3c.github.io/FileAPI/#file-section

Closes: https://github.com/bytecodealliance/StarlingMonkey/issues/155


builtins/web/file.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    if (!it.next(&item, &done)) {
+      return false;
+    }
+    if (item.isObject() && File::is_instance(&item.toObject())) {
+      instance.setObject(item.toObject());
+    }
+  }
+
+  return true;
```
tschneidereit, line 33: While this code is functionally correct, I'd like to suggest some restructuring and simplifications (note that `is_instance` has an overload that takes a `JS::Value`):
```suggestion
  instance.setNull();

  if (!(value.isObject() && it.valueIsIterable())) {
    return true;
  }

  JS::ForOfIterator it(cx);
  if (!it.init(value, JS::ForOfIterator::AllowNonIterable)) {
    return false;
  }
  JS::RootedValue item(cx);
  bool done;
  if (!it.next(&item, &done)) {
    return false;
  }
  if (File::is_instance(item)) {
    instance.set(item);
  }
```

tschneidereit (changes requested):
While I left a few inline comments on the code as written, after getting far enough in the review I realized that I find it very difficult to reason through how the code implements the specification.

For example, when trying to understand the `File` constructor, I'm not sure how it implements the log of (simplifying here!) iterating through the list of `fileBits` and appending them to the blob underlying the `File`. If I understand the code correctly, it has a special case for a sequence whose first entry is a `File` instance. If that case is hit, it copies that `File`'s contents and ignores the rest of the sequence?

But I might misunderstand the code here—that's very plausible!

Which is why I'd like to ask for something that's unfortunately a bunch of additional work: could you go through the code and add comments for which parts of the spec are implemented where? You can see examples of that in the various other builtins, in particular the ones for the `fetch` API. One that I think is quite decent is [Response::redirect](https://github.com/bytecodealliance/StarlingMonkey/blob/fb85134f84e90216b137cbe0f4ee1bbcbe552a1d/builtins/web/fetch/request-response.cpp#L2094-L2158).

I used to have a script to turn the [spec's source](https://github.com/w3c/FileAPI/blob/main/index.bs) into C++ comments, but unfortunately seem to have lost it :( In the meantime, perhaps the easiest way to quickly get decent comments to use is roughly this:
1. find the spec algorithm you want to implement/get comments for
2. select the algorithms text in your browser
3. select "view source" (this works well at least in Firefox, I haven't tried other browsers)
4. copy the source
5. paste it into an html-to-markdown converter, such as https://htmlmarkdown.com/
6. prefix all lines in the output with `//`

That should give you something that's at least not too far away from usable.

As I said, I know that I'm asking for a whole bunch of extra work here, and I'm sorry for that. I don't really know of a better way to ensure that we're actually implementing the spec as written: the WPT suite gives us decent assurances, but it's not complete and doesn't cover all corner cases.

andreiltd:
Hey @tschneidereit ! Thanks for the feedback! I can definitely add comments with the steps from the spec - that’s a good point.

I've been thinking about this implementation, and I feel like part of the issue here is that I don’t have a clear understanding of how to model inheritance using `jsapi`. According to the spec, the File API extends Blob. This patch models it using composition: embedding the `Blob` into a `File` slot and delegating methods and properties. Maybe that's a correct way, but...

I noticed that the `JS_InitClass` function allows you to pass a prototype of the parent class, which seems like something I should use here. But I couldn’t find any documentation on how to properly implement this. For example, how do I call the parent constructor from a subclass? How do I control which constructor parameters are passed to the parent?

We shouldn’t need to add a special case for something like `File([new File(["TEST"])`. This should already be covered by the `Blob` implementation since it’s equivalent to `File([new Blob(["TEST"])`. In other words, `Blob::is_instance()` should return true for both `Blob` and `File` objects.

BTW, I don't mind redoing this PR from scratch as long as we can come up with better and more idiomatic code :)

andreiltd:
Hey @tschneidereit.I refactored the code so that `File` is now a subclass of `Blob`. This simplifies the implementation significantly and, hopefully doesn’t rely on too much "black magic." I’ve also added the steps from the spec in the comments.

I will prepare a separate PR for `Blob`, adding steps from spec to the documentation.

builtins/web/file.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+JSObject *File::blob(JSObject *self) {
+  MOZ_ASSERT(is_instance(self));
+  auto blob = &JS::GetReservedSlot(self, static_cast<size_t>(Blob::Slots::Data)).toObject();
+
+  MOZ_ASSERT(blob);
```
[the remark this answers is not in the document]
andreiltd: This is now removed.

builtins/web/file.cpp
andreiltd: This is now removed.

builtins/web/blob.h
```diff
@@ -51,13 +51,19 @@ class Blob : public TraceableBuiltinImpl<Blob> {
   static const JSPropertySpec properties[];
 
   static constexpr unsigned ctor_length = 0;
-  enum Slots { Data, Type, Endings, Readers, Count };
+  enum Slots { Data, Type, Endings, Readers, Reserved1, Reserved2, Count };
```
guybedford, line 54: Since these are being reserved _by_ the FIle API, it may just make sense to directly call these `MaybeFileName` and `MaybeFileModified`, since those are their only meanings. Alternatively explicitly note they are reserved by the File API - `FileAPISlot1` and `FileAPISlot2`.

builtins/web/file.cpp
```diff
@@ -0,0 +1,199 @@
...
+          return true;
+        }
+      }
+    }
+  }
+
+  // If `lastModified` is not provided, set d to the current date and time.
+  auto now = std::chrono::system_clock::now();
+  auto ms_since_epoch =
+      std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count();
```
guybedford, line 41: We should use the hostAPI wrapper function for these - `MonotonicClock::now`.

builtins/web/file.cpp
```diff
@@ -0,0 +1,199 @@
...
+const JSPropertySpec File::properties[] = {
+    JS_PSG("name", File::name_get, JSPROP_ENUMERATE),
+    JS_PSG("lastModified", File::lastModified_get, JSPROP_ENUMERATE),
+    JS_STRING_SYM_PS(toStringTag, "File", JSPROP_READONLY),
+    JS_PS_END,
+};
+
+bool File::name_get(JSContext *cx, unsigned argc, JS::Value *vp) {
+  METHOD_HEADER(0);
+  // TODO: Change this class so that its prototype isn't an instance of the class
```
guybedford, line 81: Can you clarify what is meant by this? Is the instance prototype not supposed to be an instance of the class, or the `File.prototype` constructor `"prototype"` property?

builtins/web/blob.h
guybedford: Also I'm curious why not just have a `File` slot and have it point to a file instance here, where the File builtin has the `Name` and `Modified` slots?

andreiltd: My original goal was to use `construct`
```cpp
JS::Construct(cx, blob_ctor_val, file_ctor_val, blob_args, &self);
```
so that internally it behaved like:
```
Reflect.construct(Blob, blobArgs, File);
```

and automatically allocated an object with the File class (allowing to set up extra slots such as `name` and `lastModified`). But for some reason `JS_NewObjectForConstructor` in the `Blob` constructor always produced a Blob object regardless of the `newTarget`.

I now switched to a simpler manual approach:
- Create a File-class object directly,
- Call Blob’s initialization routine on it (for shared functionality),
- Then add the File-specific slots in the File constructor.

This keeps the “subclass” logic clean, avoids forcing Blob to know about File, and ensures we get a proper File object with all its extra slots.

andreiltd: Interestingly it also looks like using subclassing is now discouraged:

> Warning: The standard committee now holds the position that the built-in subclassing mechanism in previous spec versions is over-engineered and causes non-negligible performance and security impacts. New built-in methods consider less about subclasses, and engine implementers are [investigating whether to remove certain subclassing mechanisms](https://github.com/tc39/proposal-rm-builtin-subclassing). Consider using composition instead of inheritance when enhancing built-ins.

builtins/web/file.cpp
andreiltd: I've added the `WallClock` simple wrapper. The File implementation expects the timestamp to be the time in ms since Unix Epoch. This is what `wasi_wall_clock` API provides.

Should we also revisit the `Performance` implementation? It looks like it is using `std::chrono`.

builtins/web/file.cpp
andreiltd: Sorry, no idea. I've been carrying this comment from other classes without giving it much thought.

host-apis/wasi-0.2.0/host_api.cpp
```diff
@@ -107,6 +107,22 @@ void MonotonicClock::unsubscribe(const int32_t handle_id) {
   wasi_io_poll_pollable_drop_own(own_pollable_t{handle_id});
 }
 
+uint64_t WallClock::now() {
+  wasi_clocks_wall_clock_datetime_t time_now;
+  wasi_clocks_wall_clock_now(&time_now);
+
+  uint64_t total_nanoseconds = time_now.seconds * 1000000000ULL + time_now.nanoseconds;
```
andreiltd, line 114: This will overflow around year 2554. `¯\_(ツ)_/¯`

builtins/web/file.cpp
tschneidereit: For reasons I forgot, our builtin initialization system can't easily make it so that the prototype of instances isn't also an instance of the builtin. Which for some builtins it needs to, while for others it's not supposed to be. Hence, this `TODO` is strewn all over the place :/

builtins/web/file.cpp
```diff
@@ -122,45 +107,47 @@ bool File::lastModified_get(JSContext *cx, unsigned argc, JS::Value *vp) {
...
 
-bool File::init_last_modified(JSContext *cx, HandleObject self, HandleValue initv) {
-  bool has_last_modified = false;
-
-  JS::RootedValue init_val(cx, initv);
-  JS::RootedObject opts(cx, init_val.toObjectOrNull());
-
-  if (!opts) {
-    return true;
+bool File::init(JSContext *cx, HandleObject self, HandleValue fileBits, HandleValue fileName, HandleValue opts) {
```
tschneidereit, line 116: Can you add a doc comment with a link to the algorithm implemented here? See some of the other builtins for how I usually format those.

builtins/web/file.cpp
```diff
@@ -122,45 +107,47 @@ bool File::lastModified_get(JSContext *cx, unsigned argc, JS::Value *vp) {
...
+  //  2. F refers to the bytes byte sequence.
+  //  3. F.size is set to the number of total bytes in bytes.
+  //  4. F.name is set to n.
+  //  5. F.type is set to t.
+  //  6. F.lastModified is set to d.
+  //
+  //  Steps 2, 3 and 5 are handled by Blob. We extend the Blob by adding a `name`
+  //  and the `lastModified` properties.
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::Name), JS::StringValue(name));
+  SetReservedSlot(self, static_cast<uint32_t>(Slots::LastModified), lastModified);
```
tschneidereit, line 148: This part is a bit unfortunate: we can't store `lastModified` as a `uint32_t`, because that'd overflow very quickly. (Way way before 2038, because it includes milliseconds.)

I think the best thing to do so we don't have to step away from using a `Value` encoding is to use a `Number`: 53 bits of precision should still be plenty for our purposes here.

builtins/web/blob.h
```diff
@@ -51,13 +51,19 @@ class Blob : public TraceableBuiltinImpl<Blob> {
   static const JSPropertySpec properties[];
 
   static constexpr unsigned ctor_length = 0;
-  enum Slots { Data, Type, Endings, Readers, Count };
+  enum Slots { Data, Type, Endings, Readers, Reserved2, Reserved1, Count };
```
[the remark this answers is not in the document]
tschneidereit: heh, turns out I had left a bunch of comments quite a while ago in a draft, and those are all outdated now. GitHub simply didn't show them to me anymore though. I'll resolve them

builtins/web/blob.h
tschneidereit: Note that this recommendation is to spec authors, and doesn't apply to implementations of existing specs. Since `File` is specified as extending `Blob`, we unfortunately don't get to decide that it shouldn't do so.

andreiltd:
Hey @tschneidereit thanks for the review! Are you sure looked at the most recent version of the code? It looks like there are a bunch of comments that refer to outdated code, like Slots handling - there are now defined on File class, or using Int32 for storing `lastModified` - which uses `Number` now.

builtins/web/file.cpp
```diff
@@ -1,38 +1,72 @@
...
+#include "js/CallAndConstruct.h"
+#include "js/TypeDecls.h"
 
-  bool is_iterable = value.isObject() && it.valueIsIterable();
-  bool done;
+namespace {
 
-  if (is_iterable) {
-    JS::RootedValue item(cx);
+bool init_last_modified(JSContext *cx, HandleValue initv, MutableHandleValue rval) {
```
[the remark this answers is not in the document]
andreiltd: Will this work for `Date` object too?

andreiltd:
> Hey @tschneidereit thanks for the review! Are you sure you looked at the most recent version of the code? It looks like there are a bunch of comments that refer to outdated code, like Slots handling - there are now defined on File class, or using Int32 for storing `lastModified` - which uses `Number` now.

Anyway, I think the last commit is addressing all the suggestions :)

tschneidereit:
> Hey @tschneidereit thanks for the review! Are you sure you looked at the most recent version of the code? It looks like there are a bunch of comments that refer to outdated code, like Slots handling - there are now defined on File class, or using Int32 for storing `lastModified` - which uses `Number` now.

Yeah, sorry about that. See [my comment here](https://github.com/bytecodealliance/StarlingMonkey/pull/181#discussion_r1910387616) for how that happened.

The Int32 thing was still relevant though, because when `lastModified` wasn't provided, the timestamp was stored as an `Int32`, not a `Number`.

tschneidereit (approved):
This turned out really nice—great work!

andreiltd:
> This turned out really nice—great work!

Thanks! The SpiderMonkey API surface is sometimes overwhelming. I really appreciate all the ergonomics suggestions!

PR #183, opened 2024-12-06
fix: allow WASI adapter to be optional for hosts

guybedford:
This makes the WASI adapter optional for host-apis like Fastly's that don't use it.


tschneidereit (approved):
Makes sense

PR #185, opened 2024-12-09
fix: remove invalid event loop interest bump

guybedford:
There are two problems with this event loop interest management for the new blob StreamTask implementation:

1. I'm not seeing where non-cancelled stream tasks are getting their interest decremented
2. Event loop interest is purely about the request / response lifecycle requirements and not arbitrary immediate tracking

//cc @andreiltd

### DIFF
--- a/builtins/web/blob.cpp
+++ b/builtins/web/blob.cpp
@@ -184,7 +184,6 @@ class StreamTask final : public api::AsyncTask {
 
 public:
   explicit StreamTask(const HandleObject source, api::Engine *engine) : source_(source) {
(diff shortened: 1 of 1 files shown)


guybedford:
Strangely the WPT cases are passing here but failed for me locally. Will need some further investigation before landing.

guybedford:
Can confirm the WPT _are_ passing for me locally so it was a local build issue. In this case this should be okay to land as-is.

andreiltd (approved):

tschneidereit (approved):

Issue #182, opened 2024-12-05
URL.createObjectURL support


guybedford:
Since we now support Blobs and have incoming work on the File API, this would be nice to be supported through the request-response URL system.

PR #187, opened 2024-12-10
Implement `URL.createObjectURL` and `URL.revokeObjectURL` methods

andreiltd:
This adds `URL.createObjectURL` and `URL.revokeObjectURL` static methods to URL as well as support for blob-scheme fetch:
- https://w3c.github.io/FileAPI/#dfn-createObjectURL
- https://fetch.spec.whatwg.org/#scheme-fetch

Some noteworthy changes include:
- move uuid generation function to a separate file so that it can be used outside `crypto` component,
- split main-fetch function into two `fetch-https` and `fetch-blob` functions called based on url scheme,
- implement Range content fetch for blobs,
- add integration tests for content range fetch.

Closes: https://github.com/bytecodealliance/StarlingMonkey/issues/182

### DIFF
--- a/builtins/web/fetch/fetch-api.cpp
+++ b/builtins/web/fetch/fetch-api.cpp
@@ -47,6 +47,26 @@ std::optional<FetchScheme> scheme_from_url(const std::string_view &url) {
   }
 }
 
+// https://fetch.spec.whatwg.org/#concept-network-error
+bool network_error(JSContext *cx, HandleObject response_promise, MutableHandleValue rval) {
+  // A network error is a response whose type is "error", status is 0, status message is the empty
+  // byte sequence, header list is « », body is null, and body info is a new response body info.
+  RootedObject response_obj(cx, Response::create(cx));
+  if (!response_obj) {
+    return false;
+  }
+
+  Response::set_type(response_obj, Response::Type::Error);
+  Response::set_status(response_obj, 0);
+  Response::set_status_message_from_code(cx, response_obj, 0);
+
+  JS::RootedValue result(cx);
+  result.setObject(*response_obj);
+  JS::ResolvePromise(cx, response_promise, result);
+
+  return api::throw_error(cx, FetchErrors::FetchNetworkError);
+}
+
 bool fetch_https(JSContext *cx, HandleObject request_obj, HostString method, HostString url,
                  MutableHandleValue rval) {
   unique_ptr<host_api::HttpHeaders> headers =
@@ -115,7 +135,7 @@ bool fetch_blob(JSContext *cx, HandleObject request_obj, HostString method, Host
 
   // 2. If request's method is not `GET` or blobURLEntry is null, then return a network error.
   if (std::memcmp(method.ptr.get(), "GET", method.len) != 0) {
-    return api::throw_error(cx, FetchErrors::FetchNetworkError);
+    return network_error(cx, response_promise, rval);
   }
 
   // 3. Let requestEnvironment be the result of determining the environment given request.
@@ -131,7 +151,7 @@ bool fetch_blob(JSContext *cx, HandleObject request_obj, HostString method, Host
 
   // 8. If blob is not a Blob object, then return a network error.
   if (!blob || !Blob::is_instance(blob)) {
-    return api::throw_error(cx, FetchErrors::FetchNetworkError);
+    return network_error(cx, response_promise, rval);
   }
 
   // 9. Let response be a new response.
@@ -183,7 +203,7 @@ bool fetch_blob(JSContext *cx, HandleObject request_obj, HostString method, Host
     // 5 and 6 see extract_range function
     auto maybe_range = extract_range(range_query, full_len);
     if (!maybe_range.has_value()) {
-      return api::throw_error(cx, FetchErrors::FetchNetworkError);
+      return network_error(cx, response_promise, rval);
     }
 
     const auto [start_range, end_range] = maybe_range.value();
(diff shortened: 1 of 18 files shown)


guybedford:
Will ask @tschneidereit if he has any ideas about the UUID implementation - is there a small implementation we could inline into the codebase that avoids unnecessary code and might save some bytes? In general we want to watch binary size very carefully.

As for the request implementation it would be supporting the Blob URL everywhere we normally support a URL - so specifically `fetch(blobUrl)`.

The worker location is the base origin to use.

andreiltd:
Just for context, the size difference between the binary from main and this branch is 26KB. I'm not sure how much of this is due to the uuid crate.

guybedford:
Actually I think you could just use our existing implementation of `Crypto::random_uuid` here and then there's no need for a third-party crate.

andreiltd:
> Actually I think you could just use our existing implementation of `Crypto::random_uuid` here and then there's no need for a third-party crate.

Nice, I totally missed that :)

tests/wpt-harness/expectations/fetch/api/basic/scheme-blob.sub.any.js.json
```diff
@@ -0,0 +1,56 @@
...
+    "status": "PASS"
+  },
+  "Fetching URL.createObjectURL(empty_type_blob) is OK": {
+    "status": "PASS"
+  },
+  "Fetching URL.createObjectURL(empty_data_blob) is OK": {
+    "status": "PASS"
+  },
+  "Fetching URL.createObjectURL(invalid_type_blob) is OK": {
+    "status": "FAIL"
```
andreiltd, line 39: This test checks that blob created with `type: "invalid"` should result in empty `Content-Type` header. It assumes the blob type should me mime parsable although this assumption is not consistent with other blob tests and the spec.

As stated here: https://w3c.github.io/FileAPI/#constructorBlob

> If the [type](https://w3c.github.io/FileAPI/#dfn-BPtype) member of the [options](https://w3c.github.io/FileAPI/#dom-blob-blob-blobparts-options-options) argument is not the empty string, run the following sub-steps:
> - Let t be the [type](https://w3c.github.io/FileAPI/#dfn-BPtype) dictionary member. If t contains any characters outside the range U+0020 to U+007E, then set t to the empty string and return from these substeps.
> - Convert every character in t to [ASCII lowercase](https://infra.spec.whatwg.org/#ascii-lowercase).

cmake/compile-flags.cmake
```diff
@@ -17,7 +17,7 @@ list(APPEND CMAKE_EXE_LINKER_FLAGS
 list(JOIN CMAKE_EXE_LINKER_FLAGS " " CMAKE_EXE_LINKER_FLAGS)
 
 list(APPEND CMAKE_CXX_FLAGS
-        -std=gnu++20 -Wall -Werror -Qunused-arguments
+        -std=gnu++20 -Wall -Werror -Qunused-arguments -Wimplicit-fallthrough
```
andreiltd, line 20: Sorry for the noise here. I just spent way too much time debugging problems that were result of missing break; This just helps me to remember that `switch` is not `match`.

builtins/web/url.cpp
```diff
@@ -468,6 +482,137 @@ const JSPropertySpec URL::properties[] = {
...
+
+  auto uuid = maybe_uuid.value();
+  result.append(uuid);
+
+  RootedString url(cx, JS_NewStringCopyZ(cx, result.c_str()));
+  if (!url) {
+    return false;
+  }
+
+  if (!URL_STORE.get().put(result, obj)) {
```
andreiltd, line 570: We could probably only use `uuid` part of url as a key to save some cycles.

builtins/web/fetch/fetch-api.cpp
```diff
@@ -96,7 +99,273 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
...
+  }
+
+  rval.setObject(*response_promise);
+
+  // 1. Let blobURLEntry be request's current URL's blob URL entry.
+  RootedString blob_url(cx);
+
+  // 2. If request's method is not `GET` or blobURLEntry is null, then return a network error.
+  if (std::memcmp(method.ptr.get(), "GET", method.len) != 0) {
+    return api::throw_error(cx, FetchErrors::FetchNetworkError);
```
guybedford, line 122: I suppose this is an existing issue, but I believe fetch network errors are supposed to be DOMException::raise.

builtins/web/fetch/fetch-api.cpp
```diff
@@ -96,7 +99,273 @@ bool fetch(JSContext *cx, unsigned argc, Value *vp) {
...
+    } else {
+      // 1. If rangeStart is greater than or equal to fullLength, then return a network error.
+      if (maybe_start_range.value() > full_len) {
+        return api::throw_error(cx, FetchErrors::FetchNetworkError);
+      }
+      // 2. If rangeEnd is null or rangeEnd is greater than or equal to fullLength, then set
+      // rangeEnd to fullLength - 1.
+      start_range = maybe_start_range.value();
+      end_range = std::min(maybe_end_range.value_or(full_len - 1), full_len - 1);
+    }
```
guybedford, line 235: Could we refactor the above into a separate public function `std::tuple<size_t, size_t> Blob::extract_range(blob, std::string_view range_query)` or similar?

guybedford (changes requested):
Just a couple of points. Having some more refactoring of shared functions into blob could be useful without introducing unnecessary dependence. Otherwise would be great to land soon.

builtins/web/fetch/fetch-api.cpp
andreiltd: I think this could be done for any resource not blob only. I moved this to `fetch-utils.h` file. I hope it's fine :)

builtins/web/fetch/fetch-api.cpp
andreiltd: I tried changing this calls to `return DOMException::raise(cx, "The owls are not what they seem", "TypeError");` but then I'm getting bunch of test failures like this:

```
1:   NAME:    Fetching [PUT] URL.createObjectURL(blob) is KO
1:   MESSAGE: promise_rejects_js: function "function() { throw e }" threw object "TypeError: Invalid request method for scheme fetch" ("TypeError") expected instance of function "function TypeError() {
```

I'm not sure if it's possible to use `DOMException` the way the tests expects?

builtins/web/fetch/fetch-api.cpp
guybedford: Thanks, yes that works and will help us a lot to implement this at Fastly as we build our own request-response.cpp file implementation with Fastly-specific handling.

builtins/web/fetch/fetch-api.cpp
guybedford: That might be a deeper issue with our implementation of DOMException then. See here how `throw a TypeError` in the spec actually means throw a TypeError which is a DOM error - https://webidl.spec.whatwg.org/#exceptiondef-typeerror.

We can handle this in a follow-up of course, just wanted to mention.

guybedford (approved):
Would be great to get this landed asap. Ping @tschneidereit for final review.

builtins/web/fetch/fetch-api.cpp
andreiltd: > That all said, returning a network error looks like it should return a response? https://fetch.spec.whatwg.org/#concept-network-error

Indeed! It looks like there is a discrepancy between the tests and the spec. E.g. see this function: https://github.com/web-platform-tests/wpt/blob/master/fetch/api/basic/scheme-blob.sub.any.js#L22

It only expects `scheme-fetch` to throw  a `TypeError` on failure. I've added a `network_error` function that sets the response to the object according to the spec, but throws a `TypeError` anyway to be compliant with the tests.

guybedford:
Will aim to land this one later today unless there is further feedback.

PR #193, opened 2025-01-08
feat: support generation integer on headers objects

guybedford:
This adds support for a generation integer associated with headers mutations.

This allows for a quick check to determine if headers have been mutated since last seen by some code, useful for systems that need to store derived data based on headers (currently which we need at Fastly for cache state computations).


tschneidereit (approved):
Makes sense to me

PR #194, opened 2025-01-09
refactor: code structure

vados-cosmonic:
This PR refactors the C++ code to be stored under `src` and updates build and machinery to match.


scripts/componentize.sh
```diff
@@ -2,10 +2,10 @@
 
 #set -euo pipefail
 
-wizer="${WIZER:-@WIZER_DIR@/wizer}"
-wasm_tools="${WASM_TOOLS:-@WASM_TOOLS_BIN@}"
-weval="${WEVAL:-@WEVAL_BIN@}"
-aot=@AOT@
+wizer="${WIZER:-/home/mrman/code/work/cosmonic/forks/StarlingMonkey/deps/cpm_cache/wizer/c1b712470d9462829d63752639ff795f60d7226b/wizer/wizer}"
```
cfallin, line 5: It looks like this script got variable-substituted somehow (maybe you copied in a version from a build directory?)?

vados-cosmonic: Ah sorry, I saw this before but forgot to take it out before putting up the draft -- was working on the componentize script in another branch.

README.md
```diff
@@ -71,7 +71,7 @@ Then, the `starling-raw.wasm` module can be turned into a component with the fol
 
 ```bash
 cd cmake-build-release
-./componentize.sh -o starling.wasm
+./scripts/componentize.sh -o starling.wasm
```
andreiltd, line 74: The `script/componentize.sh` is a template that is turned into actual script after building the project. I don't think running `./scripts/componentize.sh` will work. The `<builddir>/componentize.sh` needs to be called instead.

I think we should rename the file to `script/componentize.sh.in` to highlight the fact that it is a template and shouldn't be run directly. That said, I believe the changes to README.md can be reverted.

Justfile
andreiltd, line 1: I don't have a strong opinion about this. I've seen both `justfile` and `Justfile`. Interestingly, `just` project itself call it `justfile`: https://github.com/casey/just/blob/master/justfile

vados-cosmonic: Good point -- I'll leave it as `justfile` for now -- definitely don't have a strong opinion on this

README.md
vados-cosmonic: Also did the rename to `script/componentize.sh.in` -- should be pretty clear now that it's a template -- I've also added a comment!

tschneidereit:
Can you provide more information on the motivation behind this refactor?

tschneidereit:
I'm asking because while most of it is mechanical, it does have enough moving parts that I'd want to review it in some detail, which would cost a bunch of time

vados-cosmonic:
Sorry I noted a bit [on Zulip](https://bytecodealliance.zulipchat.com/#narrow/channel/459697-StarlingMonkey/topic/Slight.20refactor.20of.20StarlingMonkey.20code.20layout.20.28code.20-.3E.20src.29/near/492654598) here, but the basic idea was that the code was a little difficult to find, and the separation between scripts that I was supposed to use versus stuff used by `cmake` was a bit confusing.

The refactor occurred to me as I was trying to change how starling was componentized

vados-cosmonic:
Closing for now -- hopefully we can revisit this maybe when some of the Rust enablement work lands!

PR #196, opened 2025-01-15
Filter out invalid header names when initializing Headers instances

### LINKED ISSUE #145: Fetch fails when guest content set `Host` header.
When the guest content sets the `Host` header, it fails with the following error

```bash
 Error: Failed to clone headers 
```

This is because' Host' is a forbidden header in the `wasi-http` spec. I believe that this was fixed a while back where the engine would remove the host header but it seems to have regressed now.

tschneidereit:
This also fixes small bugs in how requests are created in `fetch` and what `Response#type` returns. Together, these fix a bunch of WPT tests, and this PR also fixes #145.


guybedford (approved):
Nice to see the test coverage come so easily here. WinterTC compliance seems ever closer on the horizon!

PR #197, opened 2025-01-16
Move streaming reader out of `Blob` into a dedicated `BufReader` class

andreiltd:
This patch extracts the streaming reader logic from` Blob` and into a new `BufReader` class. This change removes A LOT of clutter from `Blob` while clearly separating the streaming concerns into a dedicated component.

The `BufReader` encapsulates the asynchronous task of reading data from a underlying source. It takes a `Read` function as a constructor parameter, allowing callers to specify how bytes are fetched into the provided buffer. The `Read` function must return the number of bytes `read` and indicate completion by setting `done` when there is no more data to read.

This refactoring is a preparatory work for a streaming `FormDataEncoder`, as discussed here: https://github.com/bytecodealliance/StarlingMonkey/pull/191#issuecomment-2591199953.


tests/integration/handlers.js
```diff
@@ -3,3 +3,4 @@ export { handler as performance } from './performance/performance.js';
 export { handler as crypto } from './crypto/crypto.js';
 export { handler as timers } from './timers/timers.js';
 export { handler as fetch } from './fetch/fetch.js';
+export { handler as file } from './file/file.js';
```
tschneidereit, line 6: Should this file be called something like `bufreader.js` instead, given that it doesn't really contain any file-related tests? Or are you planning on adding those in an upcoming PR? If so, seems okay to already be using this name.

tschneidereit (approved):
This all makes sense to me, and it'll be great to be able to reuse things for streaming from `FormData` instances

tests/integration/handlers.js
andreiltd: Good point, maybe `blob` is a better name? This tests buffer reading yes, but indirectly through `Blob` streaming interface - checks if multiple blob streams are handled concurrently without interfering with each other.

tschneidereit: `blob` would make sense to me, yes. I can land this PR once you've applied that change.

PR #198, opened 2025-01-18
Don't strip name section in RelWithDebInfo builds

### LINKED ISSUE #173: RelWithDebInfo debug names not working
Unfortunately, building with `RelWithDebInfo` still doesn't seem to provide debugging information properly.

A simple test of this is to run a build with `RelWithDebInfo` and then print the `starling-raw.wasm` binary with `wasm-tools print` - function names are fully lost in the process.

Resolving this is a prerequisite to getting debugging going for ComponentizeJS in https://github.com/bytecodealliance/ComponentizeJS/issues/137.

tschneidereit:
We meant to do this already, but in the move to CMake it got broken.

I hope that this fixes #173 🤞🏻


guybedford (approved):
That would be great if it does!

PR #190, opened 2024-12-20
Enable GC zealing for integration tests

andreiltd:
This is an experimental build of StarlingMonkey with GC zealing enabled. The integration tests are run with GC activated every 10 allocations and verify pre write barriers between instructions enabled.

PR #199, opened 2025-01-20
Ensure that monotonic clocks really are monotonic

tschneidereit:
Resuming wizened snapshots means that clocks don't necessarily stay monotonic. This patch overrides `clock_gettime` to address that, by adding an offset determined at the end of wizening.

@andreiltd, this should fix the assert you encountered in #190, and hopefully allow us to do better on validating our GC rooting.

### DIFF
--- a/runtime/js.cpp
+++ b/runtime/js.cpp
@@ -28,6 +28,33 @@ int main(int argc, const char *argv[]) {
   MOZ_ASSERT_UNREACHABLE("main() should not be called");
 }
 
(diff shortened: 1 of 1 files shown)


guybedford (approved):
Makes sense. If running on machines in different timezones, I wonder if we should also be doing a similar thing for the real time instead with a lower bound as opposed to an offset.

tschneidereit:
I don't think we need to do anything about the realtime clock: that is expected to vary, and in particular the timezone can change at any point even on one computer.

PR #200, opened 2025-01-21
Add implementation of `MultipartFormData` streaming encoder

andreiltd:
TODO: Add integration tests


andreiltd:
It looks like the CI failure is unrelated to this PR (I've observed the same failure on main branch). This is a problematic test:

```js
    const response = await fetch("https://http-me.glitch.me/meow?header=cat:é");
    strictEqual(response.headers.get('cat'), "é");

```

tschneidereit:
@andreiltd, is this ready to review, or are you still expecting to make changes to it?

andreiltd:
Yes, this is ready to review :)

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+  return normalize_newlines(chars);
+}
+
+// For field names and filenames for file fields, the result of the encoding must
+// be escaped by replacing any 0x0A (LF) bytes with the byte sequence
+// `%0A`, 0x0D (CR) with `%0D` and 0x22 (") with `%22`.
+//
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart-form-data
+std::optional<std::string> escape_newlines(std::string_view str) {
```
tschneidereit, line 87: Since this isn't just for newlines, it'd be good to give it a different name. Perhaps `escape_name`, given that this applies escaping of a set of characters in names?

builtins/web/form-data/form-data-encoder.h
```diff
@@ -0,0 +1,45 @@
...
+  static const JSPropertySpec properties[];
+
+  enum Slots { Form, Inner, Count };
+
+  static JSObject *form_data(JSObject *self);
+  static std::string boundary(JSObject *self);
+
+  static size_t query_length(JSContext *cx, HandleObject self);
+  static JSObject *encode_stream(JSContext *cx, HandleObject self);
+  static JSObject *create(JSContext *cx, HandleObject form_data);
```
tschneidereit, line 34: The implementation of these seems to be missing? (And in the case of `query_length` I also can't find any uses.)

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+    remainder_view_ = remainder_;
+  }
+
+  return written;
+}
+
+bool MultipartFormDataImpl::handle_entry_header(JSContext *cx, StreamContext &stream) {
+  auto entry = stream.entries->begin()[chunk_idx_];
+  auto header = fmt::memory_buffer();
+  auto name = escape_newlines(entry.name).value();
```
tschneidereit, line 261: I think this needs a null check?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+    MOZ_ASSERT(File::is_instance(entry.value));
+    RootedObject obj(cx, &entry.value.toObject());
+
+    RootedValue filename_val(cx, JS::StringValue(File::name(obj)));
+    auto filename = escape_newlines(cx, filename_val);
+
+    RootedString type_str(cx, Blob::type(obj));
+    auto type = core::encode(cx, type_str);
+
+    if (!filename || !type) {
```
tschneidereit, line 278: The `filename` check needs to happen right after the call to `escape_newlines`: if that operation errors, we don't want to run any more fallible code.

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+  auto res = host_api::Random::get_bytes(12);
+  if (auto *err = res.to_err()) {
+    return nullptr;
+  }
+
+  // Hex encode bytes to string
+  auto bytes = std::move(res.unwrap());
+  auto bytes_str = std::string_view((char *)(bytes.ptr.get()), bytes.size());
+  auto base64_str = base64::forgivingBase64Encode(bytes_str, base64::base64EncodeTable);
```
tschneidereit, line 459: Does this need a null check?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+  std::span<uint8_t> outbuf;
+  size_t read;
+  bool done;
+
+  size_t remaining() {
+    MOZ_ASSERT(outbuf.size() >= read);
+    return outbuf.size() - read;
+  }
+
+  template <typename I> size_t write(I first, I last) {
```
tschneidereit, line 143: IIUC, this method is deliberately infallible? If so, could you document that fact?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+  std::string boundary_;
+  std::string remainder_;
+  std::string_view remainder_view_;
+
+  size_t chunk_idx_;
+  size_t file_leftovers_;
+
+  bool is_draining() { return (file_leftovers_ || remainder_.size()); };
+
+  template <typename I> size_t write_and_cache_remainder(StreamContext &stream, I first, I last);
```
tschneidereit, line 176: The return value isn't ever used, it seems. Maybe remove it?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,492 @@
...
+size_t MultipartFormDataImpl::write_and_cache_remainder(StreamContext &stream, I first, I last) {
+  auto datasz = static_cast<size_t>(std::distance(first, last));
+  auto written = stream.write(first, last);
+
+  MOZ_ASSERT(written <= datasz);
+
+  auto leftover = datasz - written;
+  if (leftover > 0) {
+    MOZ_ASSERT(remainder_.empty());
+    remainder_.assign(first + written, last);
```
tschneidereit, line 251: Is this an infallible operation? It seems like it should need to allocate memory, so I'm not sure I understand how that works. And ideally, if it does allocate, we should make that fallible—or at least assert that it didn't fail and abort execution.

tschneidereit (changes requested):
Either I'm missing something in looking at the diff, or there is some code missing—see the inline comment on `form-data-encoder.h`.

One thing that I can't fully evaluate, because it'll presumably be part of the implementation of `encode_stream`: IIUC, field names, will always be first have their newlines normalized to `CRLF`, and then have those escaped. It'd be good to fold those into one operation, if possible.

Otherwise, I left a few comments and suggestions. I'm a bit concerned about allocation and general failure handling, so it'd be good to go over those aspects in some detail.

builtins/web/form-data/form-data-encoder.cpp
andreiltd: I've changed this to `escape_name`.

builtins/web/form-data/form-data-encoder.cpp
andreiltd: The function is overloaded but the version that doesn't take `JSContetext` is infallible. I can change this to separate function like `escape_name` and `escape_name_val` if we prefer explicit return value. Anyway, I've added the assertion for now.

builtins/web/form-data/form-data-encoder.cpp
andreiltd: Well, it technically can throw `bad_alloc` because it is internally calling `std::string::reserve` if needed - I guess this will abort excution.

https://cplusplus.com/reference/string/string/assign/
> A [bad_alloc](https://cplusplus.com/bad_alloc) exception is thrown if the function needs to allocate storage and fails.

builtins/web/form-data/form-data-encoder.cpp
andreiltd: This returns just a `std::string`.

andreiltd:
Hi @tschneidereit, as always, thank you so much for a through review!

> One thing that I can't fully evaluate, because it'll presumably be part of the implementation of encode_stream: IIUC, field names, will always be first have their newlines normalized to CRLF, and then have those escaped. It'd be good to fold those into one operation, if possible.

So the spec [says](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart%2Fform-data-encoding-algorithm) this:
> [For each](https://infra.spec.whatwg.org/#list-iterate) entry of entry list:
> 1. Replace every occurrence of U+000D (CR) not followed by U+000A (LF), and every occurrence of U+000A (LF) not preceded by U+000D (CR), in entry's [name](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-entry-name), by a string consisting of a U+000D (CR) and U+000A (LF).
>2. If entry's [value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-entry-value) is not a [File](https://w3c.github.io/FileAPI/#dfn-file) object, then replace every occurrence of U+000D (CR) not followed by U+000A (LF), and every occurrence of U+000A (LF) not preceded by U+000D (CR), in entry's [value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-entry-value), by a string consisting of a U+000D (CR) and U+000A (LF).

It's really hard to follow the spec closely because the implementation is state-machine-based, and a lot of actions that the spec calls for correspond to different states in the implementation and cannot be done together. For example, processing both names and values: names are part of the entry-header state, whereas values are handled in the entry-body state.

andreiltd:
I've adopted here a bunch of `wpt` tests here as an [integration test](https://github.com/bytecodealliance/StarlingMonkey/pull/200/files#diff-e3a538dcdad4e102d8172a88eb4b9c43606728620f139d2b3a06429b4c7b3bd5). Those should ensure the we serialize `FormData` according to the spec.

Unfortunately, I wasn't able to run those in `wpt` tests because of few issues:
- it looks like meta script functions are not visible from the test, e.g.: functions from `META: script=../support/send-file-formdata-helper.js` are missing when running this [test](https://github.com/web-platform-tests/wpt/blob/master/FileAPI/file/send-file-formdata.any.js)
- the wpt is hosting a bunch of python scripts and make the calls from within the tests, like this one:

```js
    const formDataText = await (await fetch(
      `/fetch/api/resources/echo-content.py`,
      {
        method: "POST",
        body: formData,
      },
    )).text();
```

This returns an empty string.

andreiltd:
> Can you say more about how you're trying to run the test?
Oh, I was just trying to run `wpt` tests directly by whitelisting them in the `tests.json` file. Those are the tests: https://github.com/web-platform-tests/wpt/tree/master/FileAPI/file

The failures that I described were from running `wpt-tests` so I ended up rewriting them, exactly as yo said,  and include in our integration tests.

> Ensuring that we retain the proper license.
That's a good point. Would `tests/integration/wpt` be a good directory for maintaining the `wpt` derived tests?

tschneidereit:
> That's a good point. Would `tests/integration/wpt` be a good directory for maintaining the `wpt` derived tests?

It would be, yes. Though if you're only doing this to work around weird issues with running WPT tests, then I would overall prefer figuring out what causes them. Can you give me steps to reproduce for the issue you were seeing with this test?

andreiltd:
I added the test like this:

```diff
diff --git a/tests/wpt-harness/tests.json b/tests/wpt-harness/tests.json
index 349baa1..fb757b9 100644
--- a/tests/wpt-harness/tests.json
+++ b/tests/wpt-harness/tests.json
@@ -175,6 +175,7 @@
   "FileAPI/blob/Blob-stream.any.js",
   "FileAPI/blob/Blob-text.any.js",
   "FileAPI/file/File-constructor.any.js",
+  "FileAPI/file/send-file-formdata.any.js",
   "hr-time/basic.any.js",
   "hr-time/idlharness.any.js",
   "hr-time/monotonic-clock.any.js",
```

and then running 

```sh
just wpt-test FileAPI/file/send-file-formdata.any.js
``` 
gives me this error:

```
1: Running test FileAPI/file/send-file-formdata.any.js
1: Sending request to http://127.0.0.1:7676/FileAPI/file/send-file-formdata.any.js
1: wasmtime stderr: stderr [0] :: Warning: Using a DEBUG build. Expect things to be SLOW.
1: wasmtime stdout: stdout [0] :: Log: running test /FileAPI/file/send-file-formdata.any.js
1: wasmtime stdout: stdout [0] :: Log: error: ReferenceError: formDataPostFileUploadTest is not defined, stack:
1: stdout [0] :: @/FileAPI/file/send-file-formdata.any.js:5:3
1: stdout [0] :: evalAllScripts@wpt-test-runner.js:5043:5
1: stdout [0] :: handleRequest@wpt-test-runner.js:5024:19
1: stdout [0] :: async*@wpt-test-runner.js:5073:54
1: stdout [0] :: 
```

andreiltd:
Just looking at how we import meta scripts [here](https://github.com/bytecodealliance/StarlingMonkey/blob/main/tests/wpt-harness/post-harness.js#L32) it looks like meta scripts are evaluated separately from the test, shouldn't the meta and the test be concatenated instead?

tschneidereit:
> Just looking at how we import meta scripts [here](https://github.com/bytecodealliance/StarlingMonkey/blob/main/tests/wpt-harness/post-harness.js?rgh-link-date=2025-02-06T20%3A20%3A11Z#L32) it looks like the meta scripts are evaluated separately from the test, shouldn't the meta and the test be concatenated instead?

I don't entirely remember why I landed on this exact way of loading META scripts. I do remember that it was non-trivial to get things working, and included this weird thing where we replace `let` and `const` with `var` bindings. If instead we could just concatenate things, that'd be nice. It's possible that the key reason was to ensure that we run the test itself without changing the lines numbers: if we concatenate everything, we can't do that anymore. But as I said, I don't entirely remember

andreiltd:
Yeah, it would be really nice to have those tests working, it's worth dozens of wpt passes. Let me know if you plan to investigate otherwise I will allocate some time to it for next week. 🙂

builtins/web/form-data/form-data-encoder.h
andreiltd: The `query_length` function is now implemented. Without setting the `Content-Length` header uploading `FormData` tests were failing.

tschneidereit:
This is looking quite excellent! 🥳 I think it's very close to being ready to land—however, before doing another in-depth review, it'd be great if you could go through the code and add a lot more links to the spec algorithms implemented. Additionally, adding comments explaining non-obvious design choices would be good, too.

In particular, explaining the boundary string code with a comment covering what you explained in the initial description in this PR would be good.

(It's very much okay to have these comments be long: we'll be grateful to have detailed explanations of all this stuff in the future!)

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,714 @@
...
+  //
+  // e.g.:
+  // This implementation: --BoundaryjXo5N4HEAXWcKrw7
+  // WebKit: ----WebKitFormBoundaryhpShnP1JqrBTVTnC
+  // Gecko:  ----geckoformboundary8c79e61efa53dc5d441481912ad86113
+  auto bytes = std::move(res.unwrap());
+  auto bytes_str = std::string_view((char *)(bytes.ptr.get()), bytes.size());
+  auto base64_str = base64::forgivingBase64Encode(bytes_str, base64::base64EncodeTable);
+
+  auto boundary = fmt::format("--Boundary{}", base64_str);
```
andreiltd, line 683: Maybe `StarlingMonkeyFormBoundary` prefix would be better?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,714 @@
...
+    return false;
+  }
+}
+
+// Computes the total size (in bytes) of the encoded multipart/form-data stream.
+//
+// Returns `std::nullopt` if any string conversion fails. This function simulates
+// the multipart/form-data encoding process without actually writing to a buffer.
+// Instead, it accumulates the total size of each encoding step.
+std::optional<size_t> MultipartFormDataImpl::query_length(JSContext* cx, const EntryList *entries) {
```
tschneidereit, line 495: Not too important, but maybe use `mozilla::Result` for this instead of `std::optional`?

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,630 @@
...
+    // Add: "Content-Disposition: form-data; name=\"\""
+    total += content_disp_len;
+    total += compute_normalized_and_escaped_len(entry.name);
+
+    if (entry.value.isString()) {
+      // Terminate the header
+      total += 2 * crlf_len;
+
+      RootedValue value_str(cx, entry.value);
+      auto value = core::encode(cx, value_str);
```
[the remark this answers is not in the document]
tschneidereit: Hmm yeah, this is pretty unfortunate. I think my preferred solution would be to store the result of the encoding here. Perhaps you could make `entry.value` a variant of `Heap<Value>, HostString`? I guess that'd also mean giving it a destructor and ensuring that that's called—so not a *great* solution as such :/

I think it might also be okay to file that as an issue to be addressed after landing this, though: nothing about this is incorrect, after all.

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,714 @@
...
+bool MultipartFormData::read(JSContext *cx, HandleObject self, std::span<uint8_t> buf, size_t start,
+                             size_t *read, bool *done) {
+  MOZ_ASSERT(is_instance(self));
+
+  if (buf.empty()) {
+    *read = 0;
+    return true;
+  }
+
+  size_t bufsz = buf.size();
```
tschneidereit, line 586: Same feedback on the name

builtins/web/form-data/form-data-encoder.cpp
tschneidereit: Not sure it makes too much of a difference, but I guess it wouldn't hurt to match other implementations' approach. Up to you though!

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,714 @@
...
+
+  return self;
+}
+
+bool MultipartFormData::init_class(JSContext *cx, JS::HandleObject global) {
+  return init_class_impl(cx, global);
+}
+
+bool MultipartFormData::constructor(JSContext *cx, unsigned argc, JS::Value *vp) {
+  MOZ_ASSERT_UNREACHABLE("No MultipartFormData Ctor builtin");
```
tschneidereit, line 700: Are you absolutely positive that this really is unreachable? Usually ctors can be reached via the prototype's (or `__proto__`s) `constructor` property. And in any case, just throwing an exception without the assert seems fine.
```suggestion
```

tschneidereit (approved):
This is fantastic work!

I left a last round of feedback, but none of that should require another look after you've addressed it, so I'm happy to merge once you let me know it's ready!

tschneidereit:
Oh, and I forgot to mention: the comments you added are really excellent and made reviewing this much easier. Thank you!

builtins/web/fetch/request-response.cpp
```diff
@@ -320,6 +325,31 @@ bool RequestOrResponse::extract_body(JSContext *cx, JS::HandleObject self,
...
+    }
+
+    RootedObject stream(cx, MultipartFormData::encode_stream(cx, encoder));
+    if (!stream) {
+      return false;
+    }
+
+    auto boundary = MultipartFormData::boundary(encoder);
+    auto type = "multipart/form-data; boundary=" + boundary;
+    host_type_str = type.c_str();
```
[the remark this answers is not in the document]
andreiltd: I tried to cleanup this code a bit. I think the problem here is that at the top of this function the `content_type` is defined as `const char *content_type` and then when we pass it to the `set_valid_if_undefined` where it is implicitly converted into a `string_view. This conversion is UB if there is no trailing `\0`.

I changed this to use a `string_view` everywhere instead. This also eliminates copying.

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -0,0 +1,714 @@
...
+      if ((i + 1) < src.size() && src[i + 1] == '\n') {
+        ++i;
+      }
+    } else if (ch == '\n') {
+      len += 3; // CR -> "%0D"
+      len += 3; // LF -> "%0A"
+    } else if (ch == '"') {
+      len += 3; // -> "%22"
+    } else {
+      len += 1;
```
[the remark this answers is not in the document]
andreiltd: I think this wouldn't be correct. For example, I think we would need something like this:

```cpp
      if ((i + 1) < src.size() && src[i + 1] == '\n') {
        // Already counted 2 for CR + LF, we want total of 6 -> add 4 => this is the same
        len += 4;
        ++i;  
      } else {
        // Single CR: already counted 1, need total of 6 -> add 5 => we need another branch here
        len += 5;
      }
```

tests/integration/formdata/formdata.js
```diff
@@ -0,0 +1,79 @@
+import { serveTest } from "../test-server.js";
```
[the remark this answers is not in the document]
andreiltd: I removed this test 🎉

builtins/web/form-data/form-data-encoder.cpp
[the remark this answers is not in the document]
andreiltd: I created an issue for this here: https://github.com/bytecodealliance/StarlingMonkey/issues/211

builtins/web/form-data/form-data-encoder.cpp
tschneidereit: yeah, the question is if any instances can possibly be reached from JS, ever. If so, it might be possible to get to the constructor from there, and then instantiate it. Which would be fine, as that'd just throw—but in that case we wouldn't want the assert here.

andreiltd: I see! OK, I removed the assert.

tschneidereit (approved):
🥳

builtins/web/form-data/form-data-encoder.cpp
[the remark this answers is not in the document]
tschneidereit: oh, good catch! Yeah, I think keeping this as-is is fine.

PR #202, opened 2025-01-22
Add `just` recipe for serving the js script

andreiltd:
This recipe will compile `starling.wasm`, componentize and serve the component with `wasmtime`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,11 @@ Once installed, navigate to the project directory and run `just` commands as nee
 just build
 ```
 
+To load a JS script during componentization and serve its output using `Wasmtime`, run:
+``` shell
+just serve <filename>.js
+```
+
 To build and run integration tests run:
 
 ``` shell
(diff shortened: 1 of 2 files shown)


guybedford:
The test service https://http-me.glitch.me/ is currently down - we might need to find a replacement, but we've alerted the team at Fastly that maintains that.

Can we add this to the README in the Just section. Alternatively it could be time to setup a DEVELOPMENT.md and make the README more end-user approachable.

andreiltd:
> The test service https://http-me.glitch.me/ is currently down - we might need to find a replacement, but we've alerted the team at Fastly that maintains that.
It looks like it's back again. Could you please poke the CI in other PR: https://github.com/bytecodealliance/StarlingMonkey/pull/200. I don't have permission to rerun CI jobs.

> Alternatively it could be time to setup a DEVELOPMENT.md and make the README more end-user approachable.
I think that's a good idea, I can prepare a separate PR for this. What would you imagine go to the DEVELOPMENT.md? Everything starting from WPT tests, and leave the first paragraph as a quickstart section?

guybedford (approved):

guybedford:
It might be worth simplifying the first section to a copy-paste case, and then having the full info in the DEVELOPMENT.md as well.

PR #203, opened 2025-01-24
stop printing request method to stderr when  is called

karthik2804:
Removes what appears to be a debug print that slipped through.


tschneidereit (approved):
Thank you!

Issue #116, opened 2024-08-27
Support FormData, Request.formData


guybedford:

stoutmusclenugget:
My team is trying to understand the timeline behind this feature. It is the last big blocker we have before going all in on Fastly Compute. Would you be able to provide an estimated timeline for this feature? Thank you!

andreiltd:
I’m going to start working on this now.

guybedford:
Just to update here - https://github.com/bytecodealliance/StarlingMonkey/pull/191 added support for the `FormData` builtin as the first part of this, but support for `request.formData` is still pending to resolve this.

andreiltd:
There are two pending PR that will resolve this:
- https://github.com/bytecodealliance/StarlingMonkey/pull/200
- https://github.com/bytecodealliance/StarlingMonkey/pull/207

I should undraft the latter shortly. I'm investigating some test failures.

PR #208, opened 2025-02-07
Implement url encoded `FormData` parser

andreiltd:
This patch adds the parser for `FormData` body that uses url encoding.

I've also attempted to clean up the `url` crate a bit: fix clippy warnings and fix potential memory leaks. Specifically, the URL API allows to create search params with `url_search_params` function. This internally allocates like this:

```rust
        url.params = Box::into_raw(Box::new(JSUrlSearchParams {
            list: url.url.query_pairs().into_owned().collect(),
            url_or_str: UrlOrString::Url(url),
        }));
```

but is never deallocated. I've added a `free_url` function that makes sure to clean up the internal state.

Closes: https://github.com/bytecodealliance/StarlingMonkey/issues/116


tschneidereit (approved):
This looks great, thank you!

PR #209, opened 2025-02-09
Fix loading of WPT META scripts

tschneidereit:
WPT tests can load helper scripts using special comments at the top. These scripts need to be loaded as though using a `<script>` tag, which doesn't create a new lexical scope, and hence causes `let` and `const` bindings to be visible in the global scope.

Previously, to simulate this behavior using `eval`, the WPT harness replaced `let` and `const` with `var`. While this worked surprisingly well, it doesn't *always* work.

This commit instead introduces an `evalScript` function available to the harness, which evaluates in a way more closely matching that of script tags.

### DIFF
--- a/justfile
+++ b/justfile
@@ -72,8 +72,8 @@ wpt-test filter="": wpt-build
 
 # Update web platform test expectations
 [group('wpt')]
-wpt-update: wpt-build
-    WPT_FLAGS="--update-expectations" ctest --test-dir {{ builddir }} -R wpt --verbose
+wpt-update filter="": wpt-build
+    WPT_FLAGS="--update-expectations" WPT_FILTER={{ filter }} ctest --test-dir {{ builddir }} -R wpt --verbose
 
 # Run wpt server
 [group('wpt')]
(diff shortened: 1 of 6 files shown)


tschneidereit:
@andreiltd this should hopefully make the tests we talked about in #200 work

andreiltd:
Hey @tschneidereit. That's a great improvement, thanks! I gave it go on the `form-data` branch and indeed the `Not Defined` error is gone. 🎉  However I'm getting this error when I try to run the tests:

```
1: WPT server stderr: 127.0.0.1 - - [09/Feb/2025 10:31:49] code 400, message Bad request syntax ('1D9')
1: 127.0.0.1 - - [09/Feb/2025 10:31:49] "1D9" 400 -
```

I will need to take a closer look if that's linked to `FormData` changes or something else although I can see the other tests are failing in a similar way. For example:

```
1: Running test fetch/api/redirect/redirect-upload.h2.any.js
1: Sending request to http://127.0.0.1:7676/fetch/api/redirect/redirect-upload.h2.any.js
1: WPT server stderr: 127.0.0.1 - - [09/Feb/2025 10:31:30] code 400, message Bad request syntax ('4')
1: 127.0.0.1 - - [09/Feb/2025 10:31:30] "4" 400 -
```

andreiltd:
I think I figured this out. The `400` error is thrown when I'm not setting the `Content-Length` header for the `Request` with `ReadableStream` body. After fixing the tests are passing 🎉 .

PR #210, opened 2025-02-10
Don't detach chunks of streaming outgoing request bodies

tschneidereit:
Detachment must only happen for streaming outgoing response bodies, duh!


tschneidereit:
One would think that this would be covered by something in the WPT suite—but one would be wrong: I couldn't find any tests that would cover any of this, unfortunately.

tschneidereit:
@andreiltd could you take a look at this and let me know if anything jumps out at you?

tschneidereit:
I discussed this with @guybedford, and we're both comfortable merging it without Guy giving a formal review

PR #212, opened 2025-02-16
Change Rust integration to emit a single static library

tschneidereit:
… instead of compiling multiple crates to static libs and trying to link all of them into the final binary.

This also replaces linking SpiderMonkey's Rust static library with including the required crates into our own library, achieving an end result of having a single static Rust library to link.

And finally, it enables thin-LTO for all libraries: this is in general useful, but even more relevant with more Rust code.


tschneidereit:
@andreiltd it'd be great if you could review this!

crates/rust-multipart/Cargo.toml
```diff
@@ -2,9 +2,6 @@
 resolver = "2"
 members = ["fuzz"]
 
-[lib]
```
andreiltd, line 5: Should we also strip following from this file?

```toml
[profile.release]
lto = true
panic = 'abort'

[profile.dev]
lto = true
panic = 'abort'
```

andreiltd (approved):
Unexpected wpt passes are indeed unexpected :)

tschneidereit:
> Unexpected wpt passes are indeed unexpected :)

Initially yes. Took some digging, but I found the cause: the `rust-url` crate has a lock file, which we're now not using anymore. And so the `url` crate was updated from `2.5.2` to `2.5.4`.

So far, so good. However, the question is if we need a lock file. And if so, how do we even go about getting one? I guess we could keep one in the template crate and copy it over. If it doesn't cover everything, it'll be added to during the build, which seems okay.

andreiltd:
Would it be possible to have a `workspace` Cargo.toml in the project root directory that `cmake` will pick up and build everything to a single library, and then maintain only a top level lock file?

andreiltd (approved):

PR #217, opened 2025-02-19
Log warnings for rejections left unhandled after FetchEvent handler completes

tschneidereit:
This logs all unhandled rejections to `stderr`, because they're pretty universally bugs to be addressed. Additionally, it logs a separate warning about rejection of promises passed to `FetchEvent#waitUntil`, for the same reason.


radu-matei:
This is incredibly helpful. Thank you!!

andreiltd (approved):

PR #218, opened 2025-02-20
Add platform support for JS content debugging

tschneidereit:
This PR introduces platform support for debugging JavaScript. "Platform support" here means that the PR doesn't contain an implementation of an actual debugger. Instead, it adds infrastructure to StarlingMonkey to initiate a debugging session over a TCP socket connection, and to load a JS implementation of an actual debugger via that connection.

The debugger itself works, but isn't quite ready for prime-time, so I'll open it as a draft PR, building on this one.

The debugger requires sort of a double opt-in: StarlingMonkey has to be started with the `-d/--enable-script-debugging` option set, and a port to connect to needs to be provided via the `DEBUGGER_PORT` env var. That ensures that StarlingMonkey will not try to read env vars at runtime without a prior opt-in, and hence will guarantee that it'll continue working in environments that stub out `wasi::environment`.

@andreiltd I think this is in a state where we can land it, and iterate in-tree. The nice thing about doing so would be that from then on, iterating on the debugger implementation doesn't require custom builds of StarlingMonkey anymore, which is why I'd like to land this now, instead of waiting until the debugger has reached maturity, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,12 @@ include("openssl")
 include("${HOST_API}/host_api.cmake")
 include("build-crates")
 
+option(ENABLE_JS_DEBUGGER "Enable support for debugging content via a socket connection" ON)
+
 add_library(extension_api INTERFACE include/extension-api.h runtime/encode.h runtime/decode.h)
+if (ENABLE_JS_DEBUGGER)
+    target_compile_definitions(extension_api INTERFACE ENABLE_JS_DEBUGGER)
+endif()
 target_link_libraries(extension_api INTERFACE rust-url spidermonkey)
 target_include_directories(extension_api INTERFACE include deps/include runtime)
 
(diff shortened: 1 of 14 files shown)


andreiltd (approved):
This is great, and it makes a lot of sense to land before the DAP protocol implementation. 

The only suggestion I have is to make the debugger feature as difficult as possible to accidentally include in prod. IIUC double opt-in is a runtime check? I'm probably a bit paranoid but I would prefer stripping all the debugger code at built time.  A single `cmake` switch seems a bit risky. Maybe we can restrict the feature to debug builds only? Or, if we allow it in release builds, we should at least include a FAT warning indicating "Hey, you have a backdoor".

tschneidereit:
@andreiltd I think hiding this behind a compile time flag would sort of deafeat the purpose: it'd make it very hard to use this for debugging content. In particular for future scenarios such as debugging content in remote environments, where it's even more important to debug in the runtime you're deploying with.

I also don't think there's any actual backdoor here: you have to first pass a cmd option during wizening, and then you have to set the `DEBUGGER_PORT` env var at runtime. I can't think of any scenario in which either would happen by accident, or maliciously, and you don't have much bigger problems than this :)

tschneidereit:
@andreiltd do you find that sufficiently convincing to land this as-is? Otherwise, I'd be happy to chat about it

PR #219, opened 2025-02-20
Add a VSCode extension for debugging JS running in StarlingMonkey

tschneidereit:
This commit introduces a substantial new feature: debugging support via the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/).

While for now the debugger is experimental and not quite ready for production use, it already covers a decent amount of functionality: it supports breakpoints, single-stepping into and over function calls, stack and (local and global) scope inspection, and modifying locals and object properties.

Building on the platform support added in #218, the extension has two key components:
1. a DAP implementation as a Node.js script that starts StarlingMonkey, starts a server for the debug session, and translates between the messages StarlingMonkey sends/receives and the DAP
2. a VS Code extension providing launch configurations and settings for the debugger, and initiates the debug session


squillace:
FWIW, I'd expect a "server" process to continue in debug mode until directly stopped, rather than have debug mode stop after each request. Sometimes it takes several iterations of walking through to realize precisely what the problem is.

andreiltd:
Hey @tschneidereit I'm curious about your thoughts on how do you envision running the debugger for scripts that don't target the `incoming-handler`? 

FWICT, we currently initialize the debugger on incoming request. I'm wondering -- would it be possible to initialize it during engine startup instead (wizer resume perhaps) to allow debug components regardless of whether they have an incoming HTTP handler installed?

squillace:
Yes, probably we should have a way to a) configure whether debug is unconditionally started or not and b) figure out whether the specific solution there would require specifying an api call to start on? Wide open as to what the most flexible method would be here. We have some engineering time to burn here as well....

tschneidereit:
> configure whether debug is unconditionally started or not

I don't think we need this part, as in a way it doesn't entirely make sense: we can only start debugging if a socket port to connect to has been provided during `wasmtime serve` as part of telling StarlingMonkey to enable debugging.

Basically, I think all that's needed is to add [this call](https://github.com/bytecodealliance/StarlingMonkey/blob/3dfbcb4e724168b9a9f80b0403d50d19e95b5be7/builtins/web/fetch/fetch_event.cpp#L547C3-L547C55) to the handler for incoming calls. I can't immediately see a reason why this couldn't be moved to [here](https://github.com/bytecodealliance/StarlingMonkey/blob/3dfbcb4e724168b9a9f80b0403d50d19e95b5be7/host-apis/wasi-0.2.0/host_api.cpp#L1028), which would take it out of the builtins implementation and into the runtime. That's probably the better place for it, and something we could readily document

PR #220, opened 2025-02-25
Implement `EventTarget` and `Event` builtin

andreiltd:


guybedford:
Great to see this. Strictly speaking `FetchEvent` is supposed to be a subclass of this `Event`. Not sure how practical that will be though, but I think that's the WinterTC expectation.

andreiltd:
> Great to see this. Strictly speaking `FetchEvent` is supposed to be a subclass of this `Event`. Not sure how practical that will be though, but I think that's the WinterTC expectation.

Yeah, I think it should be doable similar to how we made `File` extending `Blob`. I will look into this as well when I have a moment. I think the other nice to have event would be `CustomEvent`.

andreiltd:
> Strictly speaking FetchEvent is supposed to be a subclass of this Event.

So this is now working as expected, for instance this code shows `FetchEvent` using `Event` interface (event.type).

```js
async function handler(event) {
  console.log(event.type);
}

addEventListener("fetch", (event) =>
  event.respondWith(
    (async () => {
      await handler(event);
      return new Response(`Success`);
    })(),
  ),
);
```

builtins/web/event/global-event-target.cpp
```diff
@@ -0,0 +1,83 @@
...
+    return false;
+  }
+
+  RootedValue type(cx, args.get(0));
+  RootedValue callback(cx, args.get(1));
+  RootedValue opts(cx, args.get(2));
+
+  args.rval().setUndefined();
+
+  std::ignore = HANDLERS->append(&callback.toObject());
```
andreiltd, line 29: There is something odd happening when I register global handlers within GLOBAL_EVENT_TARGET. This line should not be needed, but when I remove it, the callback values stored in the global event handler get invalidated. This is very surprising because I am essentially following the tracing pattern from the [SpiderMonkey embedding examples](https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/8474c9fefba9d89e1ab74efbe238fe7877d473ad/examples/tracing.cpp#L155) for custom objects almost exactly. Yet, if I remove this line, which merely references the callbacks in a global vector, the callbacks in global `EventTarget` get invalidated. 😕 

This only happens for the global instance, if I create a local `EventTarget` and add listeners to it everything works as expected.

guybedford (approved):

include/builtin.h
```diff
@@ -180,6 +180,9 @@ template <typename Impl> class BuiltinImpl {
   static constexpr JSClassOps class_ops{};
   static constexpr uint32_t class_flags = 0;
 
+  // A runtime registry for subclass JSClass pointers.
+  static inline std::vector<const JSClass *> registry;
```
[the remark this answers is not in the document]
andreiltd: This one is tricky. This has to be shared across all the instances thus `static`. For example `Event::registry` will look like this:

```cpp
Event::registry = {CustomEvent, FetchEvent}
```

This way `Event::is_instance` can check if object can be used  as `Event`. 

 I'm not aware of mechanism that would allow to define shared across all instances slot? I also tried very hard to make this `constexpr`, which should be possible, but ended up with circular includes which break the compilation.

builtins/web/fetch/fetch_event.cpp
```diff
@@ -14,44 +16,41 @@
...
+  auto count = JS::GetReservedSlot(self, FetchEvent::Slots::PendingPromiseCount).toInt32();
   count++;
   MOZ_ASSERT(count > 0);
   if (count == 1) {
     ENGINE->incr_event_loop_interest();
   }
-  JS::SetReservedSlot(self, static_cast<uint32_t>(FetchEvent::Slots::PendingPromiseCount),
-                      JS::Int32Value(count));
+
+  JS::SetReservedSlot(self, FetchEvent::Slots::PendingPromiseCount, JS::Int32Value(count));
```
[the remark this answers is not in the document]
guybedford: Ahh right, of course that makes sense. Wonder if @tschneidereit has an opinion here.

include/builtin.h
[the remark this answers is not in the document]
guybedford: Could it be a singular static managed through the macro process?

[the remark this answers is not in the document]
andreiltd: Could you give an example how this would look like?

The circular dependency problem I get comes from:

```cpp
// custom-event.h

#include "event.h"

  static constexpr int ParentSlots = Event::Slots::Count;
  enum Slots { Detail = ParentSlots, Count };
```

```cpp
// event.h
#include "custom-event.h"

  static constexpr registry = [&CustomEvent::class_];
```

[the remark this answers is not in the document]
guybedford: Another mechanism here might be to define it as instance data, even if that data is the same on every instance through its constructor?

builtins/web/event/global-event-target.cpp
andreiltd: I was finally able to track this down. The results are very interesting. It turned out that we had a flaw in the implementation of builtins. The issue was related to static member inheritance in C++ templates.

For `FinalizableBuiltinImpl` and `TraceableBuiltinImpl` we had something like this:

```cpp
template <typename Impl> class FinalizableBuiltinImpl : public BuiltinImpl<Impl> {
  static constexpr JSClassOps class_ops{
      nullptr,        // addProperty
      nullptr,        // delProperty
      nullptr,        // enumerate
      nullptr,        // newEnumerate
      nullptr,        // resolve
      nullptr,        // mayResolve
      Impl::finalize, // finalize
      nullptr,        // call
      nullptr,        // construct
      nullptr,        // trace
  };
  static constexpr uint32_t class_flags = JSCLASS_BACKGROUND_FINALIZE;
};
```

Problem is that it does not shadow the `BuiltinImpl::class_ops`. So even if we declare the builtin as:

```cpp
class MyBuiltin: FinalizableBuiltinImpl<MyBuiltin> {
}
```

the `JSClass` was still initialized with `ops` from `BuiltinImpl` which doesn't define finalize/trace. As a result we never called `finalize` nor `trace` function and the lack of latter was causing all sort of problems.

I rewrote the `Finalizable` and `Traceable` to template policy pattern and everything works as expected.

include/builtin.h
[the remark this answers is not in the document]
andreiltd: I rewrote this to encapsulate registry in the function. Let me know if this work for you. 🙂

builtins/web/event/global-event-target.cpp
andreiltd: BTW, I was able to get a lot of help on this from [`SpiderMonkey` matrix](https://matrix.to/#/!CZEbuMfGYVdQMIBKeP:mozilla.org/$LOl2q0eGZazAS64YJvj41lfe9t5ryuVCIs-5ZPfH_mY?via=mozilla.org&via=igalia.com&via=matrix.org)

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+}
+
+
+// https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke
+bool EventTarget::inner_invoke(JSContext *cx, HandleObject event, JS::HandleVector<EventListener> list, bool *found) {
+  // 1. Let found be false.
+  *found = false;
+
+  // 2. For each listener of listeners, whose removed is false:
+  for (auto &listener : list) {
```
tschneidereit, line 405: If I read things correctly, then exceptions thrown while invoking any individual listener shouldn't abort the loop, but instead keep it running. The easiest way to do that might be to move the entire loop body into its own function, except for the substeps of step 11; the error reporting.

I'm a bit surprised that this doesn't seem to be covered by WPT tests. Unless my reading is wrong, and aborting is correct here. Can you maybe verify this with a new test that you also (manually, in browser devtools) run in a browser?

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+    // 5. If listener's once is true, then remove an event listener given event's
+    //  currentTarget attribute value and listener.
+    if (listener.once) {
+      auto current_target = Event::current_target(event);
+      MOZ_ASSERT(is_instance(current_target));
+
+      auto target_list = listeners(current_target);
+      MOZ_ASSERT(target_list);
+
+      auto it = std::find_if(target_list->begin(), target_list->end(), [&](const auto &other) {
```
tschneidereit, line 437: I think it'd be great if we could skip this here. Instead, just set the `removed` flag on the listener here, perhaps add a `bool listeners_removed` outparam to this function, and then filter the target's listeners list once after the loop. It'd require changing `list` to `JS::HandleVector<&EventListener>`, but that seems like a good change to do regardless: no need to actually copy all those listener instances.

Do you think that makes sense?

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+
+    RootedValue rval(cx);
+    RootedValueArray<1> args(cx);
+    args[0].setObject(*event);
+
+    if (JS::IsCallable(callback_obj)) {
+      auto global = engine->global();
+      if (!JS::Call(cx, global, callback_val, args, &rval)) {
+        auto msg = "dispatching event " + listener.type;
+        engine->dump_pending_exception(msg.c_str());
```
tschneidereit, line 476: As mentioned above, this should all be factored out to happen for all problems in the loop, and shouldn't abort the loop itself.

Also, can you file an issue for adding support for the [report an exception](https://html.spec.whatwg.org/multipage/webappapis.html#report-an-exception) concept, and add a TODO linking to it here?

builtins/web/fetch/fetch_event.cpp
```diff
@@ -469,63 +455,15 @@ bool FetchEvent::response_started(JSObject *self) {
...
 static void dispatch_fetch_event(HandleObject event, double *total_compute) {
   MOZ_ASSERT(FetchEvent::is_instance(event));
-  // auto pre_handler = system_clock::now();
 
-  RootedValue result(ENGINE->cx());
   RootedValue event_val(ENGINE->cx(), JS::ObjectValue(*event));
-  HandleValueArray argsv = HandleValueArray(event_val);
-  RootedValue handler(ENGINE->cx());
   RootedValue rval(ENGINE->cx());
+  RootedObject event_target(ENGINE->cx(), global_event_target());
```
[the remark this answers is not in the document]
andreiltd: The `global_event_target` returns just that, the pointer to a global instance of `EventTarget`. The global event target is also installing `addEventListener`, `removeEventListener` methods on `global` object. Those are used as delegates to add/remove listeners form global `EventTarget`.

I'm not sure what the other way would be of having global functions `addEventListener` etc?

builtins/web/event/event-target.cpp
andreiltd: I think the problem is that there is nothing in the spec that would prevent a listener to modify the `EventTarget` (.e.g. removing itself from the target) during the dispatch and I believe this is why the spec explicitly calls for cloning the listeners. I think holding onto listener references during dispatch is not sound but moving removals to after the dispatch sounds like a good optimization - we can mark cloned listeners as removed and do actual removing outside dispatch loop. WDYT?

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+  if (!val.isUndefined()) {
+    passive.setBoolean(JS::ToBoolean(val));
+  }
+
+  // - If options["signal"] exists, then set signal to options["signal"].
+  if (!JS_GetProperty(cx, obj, "signal", &val)) {
+    return false;
+  }
+  if (val.isObject()) { // TODO: check if is instance of AbortSignal
+    passive.set(val);
```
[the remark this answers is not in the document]
andreiltd: Nice catch!

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+    // 1. If event's type attribute value is not listener's type, then continue.
+    RootedString type_str(cx, Event::type(event));
+    auto event_type = core::encode(cx, type_str);
+    if (!event_type) {
+      return false;
+    }
+
+    if (listener.type != std::string_view(event_type)) {
+      continue;
+    }
```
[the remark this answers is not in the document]
andreiltd: D'oh!

builtins/web/event/event-target.cpp
andreiltd: Indeed the loop should not be aborted. I've confirmed this in a browser and added integration test.

builtins/web/event/event-target.h
```diff
@@ -0,0 +1,79 @@
...
+
+  JS::Heap<JS::Value> callback;
+  JS::Heap<JS::Value> signal;
+
+  std::string type;
+
+  bool passive;
+  bool capture;
+  bool once;
+  bool removed;
```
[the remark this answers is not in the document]
andreiltd: I just did a test and replaced the bools with `unit8_t flags` and in both cases the `sizeof` of instance returns 32 bytes.

include/builtin.h
```diff
@@ -202,10 +261,23 @@ template <typename Impl> class BuiltinImpl {
...
+  static bool is_subclass(const JSObject *obj) {
+    if (obj == nullptr)
+      return false;
+
+    const JSClass *cls = JS::GetClass(obj);
+    const auto &registry = get_registry();
+
+    return std::any_of(registry.begin(), registry.end(),
+                       [cls](const auto *reg) { return reg == cls; });
+  }
```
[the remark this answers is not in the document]
tschneidereit: `is_instance` is called at least once in every builtin method, via the chain `METHOD_HEADER->check_receiver->Impl::is_instance`

builtins/web/event/event-target.cpp
tschneidereit: Looking at the _note_ under step 6 of https://dom.spec.whatwg.org/#concept-event-listener-invoke, it seems like we actively have to make the `EventListener` entries be references? If, during the `invoke` operation, a listener is removed, its `removed` flag should be set, and the listener not be invoked. But with the current implementation, we won't see the flag here, because all the entries are copies.

tschneidereit: more generally, as mentioned in that same note in the spec, the goal of the list copy is to prevent added listeners from running (presumably because that could result in infinite loops), not preventing absolutely any modifications to the entries

builtins/web/event/event-target.h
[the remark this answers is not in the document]
tschneidereit: makes sense, thank you for checking!

builtins/web/event/event-target.cpp
```diff
@@ -0,0 +1,557 @@
...
+    return true;
+  }
+
+  // Otherwise:
+  // - Return options["capture"].
+  if (opts.isObject()) {
+    JS::RootedObject obj(cx, &opts.toObject());
+    JS::RootedValue val(cx);
+
+    if (!JS_GetProperty(cx, obj, "capture", &val)) {
```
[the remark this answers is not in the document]
tschneidereit: oh right, I had forgotten that the Property-related APIs don't have straightforward overloads for JSStrings. You can use `JS_StringToId` and then `JS_GetPropertyById`, but let's not hold up this PR on that

builtins/web/event/event-target.cpp
andreiltd: That might be correct behavior, yes. The technical difficulty I mentioned stems from maintaining valid references during dispatch. Imagine code like this:

```cpp
std::vector<EventListener> listeners = EventTarget::listeners();
std::vector<&EventListener> =  cloned();  // derived from listeners

for listener in cloned {
    dispatch(| | {
        add_new_listener();
        // adding new listeners triggers reallocate of `listeners`
        // continuing the loop is UB
    });
}
```

I guess this could be solved by having `std::vector<shared_ptr<EventListener>>` in `EventTarget` instead, with the performance penalty: each listener is now separate allocation so iterating is slower due to potential cache misses and no prefetch.

include/builtin.h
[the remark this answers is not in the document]
andreiltd: > Can we perhaps add a static bool has_subclasses that gets set the first time a subclass is registered?

The `is_subclass` is only ever called through `is_instance` function, where the check looks like this:

```cpp
return (obj != nullptr) && (JS::GetClass(obj) == &class_ || is_subclass(obj));
```

Probably 97% of code will not attempt checking for subclass as the first condition will be true. If the first condition is false than it is a subclass, or else an error.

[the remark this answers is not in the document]
tschneidereit: ah, that's a good point. Once we make more things event dispatchers, that will increase a bit, but yeah, it's less of a concern than I thought. Let's keep this as-is for now!

builtins/web/event/event-target.cpp
andreiltd: OK good! I will try to rework this and add tests.

andreiltd: The "Once listener should be invoked only once even if the event is nested" test is now failing, but that's for tomorrow 🙂

andreiltd: OK so the problem is that if we defer removal of "once" listener to after the dispatch loop instead of removing them immediately, the listener dispatch function cannot re-add the handler because it already exists. There is a specific WPT test for this.

In theory we could still delay the removal if  we have an additional check in `addEventListener` function for: if listener exists but it's marked for removal then add it anyway. I'm happy to do that if we think that the potential performance gains justify diverging from spec.

andreiltd: Sorry I don't follow 😅  do you mean something like:

```cpp
  for (auto &listener : list) {
    ...
    if (listener->once) {
       listener->removed = true;
       // move to the end
    }
    
    // dispatch
  }
  
   // defer removal
   target_list->eraseIf([&listener](const auto &other) { return listener->removed; });
```

tschneidereit: Not quite, but close. IIUC, the problem is that it's possible to re-add a `once` listener during dispatching, in which case it needs to be kept in the list, but moved to the end, right? If so, I'd imagine something like this:
```cpp
  for (auto &listener : list) {
    ...
    if (listener->once) {
       listener->removed = true;
    }
    
    // dispatch
  }
  
   // defer removal
   target_list->eraseIf([&listener](const auto &other) { return listener->removed; });

// in add_listener:
  if (found(listener) && listener.removed) {
    listener.removed = false;
    // move to the end
  }
```

So if `add_listener` is calling during dispatching, we reset the `removed` flag, but also move the listener to the end, because it shouldn't keep its previous position in the list.

andreiltd: Ah yes, sure! That makes  sense, let me try this.

include/builtin.h
```diff
@@ -262,18 +258,11 @@ template <typename Impl, typename ClassPolicy = DefaultClassPolicy> class Builti
...
   static bool is_subclass(const JSObject *obj) {
-    if (obj == nullptr)
-      return false;
-
-    const JSClass *cls = JS::GetClass(obj);
-    const auto &registry = get_registry();
-
-    return std::any_of(registry.begin(), registry.end(),
-                       [cls](const auto *reg) { return reg == cls; });
+    return std::ranges::any_of(get_registry(), [&](auto *cls) { return cls == JS::GetClass(obj); });
```
tschneidereit, line 265: Is there any reason not to use `get_registry().contains()`? As far as I can tell, `any_of` still has to do a linear search, since AFAICT the condition has to be tested for each item.

builtins/web/event/event-target.cpp
```diff
@@ -357,25 +390,25 @@ bool EventTarget::invoke_listeners(JSContext *cx, HandleObject target, HandleObj
...
+  if (Event::has_flag(event, EventFlag::StopPropagation)) {
     return true;
   }
 
   // 5. Initialize event's currentTarget attribute to struct's invocation target.
   Event::set_current_target(event, target);
   // 6. Let listeners be a clone of event's currentTarget attribute value's event listener list.
   auto list = listeners(target);
-  JS::RootedVector<EventListener> list_clone(cx);
+  JS::RootedVector<ListenerRef> list_clone(cx);
```
tschneidereit, line 411: Very much getting into the weeds of performance optimization here, but using `list_clone.reserve(listeners.length())` would ensure that the list doesn't have to be resized during cloning.

builtins/web/event/event-target.cpp
```diff
@@ -471,35 +497,44 @@ bool EventTarget::inner_invoke(JSContext *cx, HandleObject event, JS::HandleVect
...
+      JS::Call(cx, callback_val, handle_fn, args, &rval);
+    }
 
-      if (!JS::Call(cx, callback_val, handle_fn, args, &rval)) {
-        auto msg = "dispatching event " + listener.type;
-        engine->dump_pending_exception(msg.c_str());
-        return false;
-      }
+    if (JS_IsExceptionPending(cx)) {
+      // TODO: report an excecption as in spec:
```
tschneidereit, line 510:
```suggestion
      // TODO: report an exception as in spec:
```

tschneidereit (approved):
I am so, so sorry for taking so long with this review! :(

This is in great shape, with only a handful of comments below. Not all of them need addressing, but I think at least one is a correctness issue.

Still, with that addressed, this will be ready to go!

builtins/web/event/event-target.cpp
```diff
@@ -220,22 +243,27 @@ bool EventTarget::add_listener(JSContext *cx, HandleObject self, HandleValue typ
...
+    listener->once = once;
+    listener->removed = false;
 
     list->append(listener);
+  } else if((*it)->removed) {
+    // if existing listener was marked for removal, then move it to the end of the list
+    // and update its removed flag. This is done to ensure that the order of listeners.
+    (*it)->removed = false;
+    list->erase(it);
+    list->append(*it);
```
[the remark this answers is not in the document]
andreiltd: That's a great catch. I updated the code and added integration test.

PR #222, opened 2025-03-03
Add support for running initialization scripts before content

tschneidereit:
This adds the ability to run  a script in a separate global before content gets run. This code has access to additional abilities, notably including the `defineBuiltinModule` function.

Building on this will allow us to change ComponentizeJS to support content using imports from `wasi:package-name` module names without rewriting the code first. That in turn will enable support for non-bundled content code.

### DIFF
--- a/componentize.sh.in
+++ b/componentize.sh.in
@@ -9,9 +9,11 @@ aot=@AOT@
 preopen_dir="${PREOPEN_DIR:-}"
 
 usage() {
-  echo "Usage: $(basename "$0")  [--verbose] [--legacy-script] [input.js] [-o output.wasm]"
+  echo "Usage: $(basename "$0")  [--verbose] [-i,--initializer-script-path] [--legacy-script] [input.js] [-o output.wasm]"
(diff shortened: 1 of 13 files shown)


tschneidereit:
@guybedford looking for your feedback here, because my hope is to build on this in ComponentizeJS to get rid of the content preprocessing, and be able to directly use `wasi:` imports.

tschneidereit:
I'm very unsure what's going on with the failing tests: whether that's an infra-issue or something else. It's certainly not related to the code changes in this PR. Will investigate

andreiltd:
> I'm very unsure what's going on with the failing tests: whether that's an infra-issue or something else. It's certainly not related to the code changes in this PR. Will investigate

This perhaps: https://github.com/rust-lang/rustup/issues/4211 ?

tschneidereit:
> This perhaps: [rust-lang/rustup#4211](https://github.com/rust-lang/rustup/issues/4211) ?

It was that indeed, yes—fixed over in #212.

tschneidereit:
@guybedford gave a verbal r+ to this in the BA JS SIG call earlier today.

PR #225, opened 2025-03-10
docs: Add Riza to the list of production adopters

andrewmbenton:
We've been happily running StarlingMonkey in production for quite awhile, and didn't realize this page existed!


andreiltd (approved):
Thank you!

PR #231, opened 2025-03-19
Support stripping a prefix from all file paths

tschneidereit:
This introduces the CLI option `--strip-path-prefix prefix` and strips the provided `prefix` from all paths as seen by content, in stack traces, and the debugger.

Previously, we unilaterally stripped the directory part of the top-level script's path instead. This new approach is better in two ways:
1. it works for files that are in ancestor or sibling directories to the top-level script
2. it enables using an ancestor directory as the root

The second of these advantages can be seen in the diffs to the test expectations files: the stacks for those are more informative, because they now include the directory of the actual test, since we're only stripping the `[StarlingMonkey root]/tests` part of the path.

### DIFF
--- a/componentize.sh.in
+++ b/componentize.sh.in
@@ -9,11 +9,12 @@ aot=@AOT@
 preopen_dir="${PREOPEN_DIR:-}"
 
 usage() {
-  echo "Usage: $(basename "$0")  [--verbose] [-i,--initializer-script-path] [--legacy-script] [input.js] [-o output.wasm]"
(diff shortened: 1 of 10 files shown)


guybedford:
Could we define a concept of base path? Or can we run Wizer such that the project root is `/`? It's always unfortunate to spill local paths into production systems IMO.

tschneidereit:
@guybedford agreed that that is much better, so I did that in this new version of this PR. Turns out without that kind of approach it was hard to make the e2e tests pass again anyway :)

guybedford (approved):

Issue #188, opened 2024-12-10
Node.js builtins


guybedford:
Tracking issue for supporting Node.js builtins (`import fs from node:fs`, ...), where each builtin can be individually turned on or off.

We should implement this layer in C++ (or Rust), and try to build optimized implementations.

yoshuawuyts:
I was wondering: Node.js compat is something I assume other WinterCG implementations would be interested in that too? Is there any work happening on a JS-based shim already? If not, could that potentially serve as a starting point?

guybedford:
A JS-based shim could well be a starting point, although in terms of that being a development blocker often a simple build step can also be used to build in Node.js compatibility just as well.

If someone wanted to work on an entirely inlined JS approach to builtins we could certainly accept those PRs as a start though.

tschneidereit:
There are JS-based shims, including browserify, yes. The biggest issue for us is that these often are built in a way that assumes that JS is faster to execute than it is in our environment. Because of that, I think it'd be unfortunate if we invested a large amount of energy into a JS based approach, instead of doing something native

smndtrl:
I wanted to add `async_hooks`/AsyncContext as a interesting candidate. I was playing around getting Next.js routes to execute as components which worked well for simple API routes but as soon as it gets a bit more complex like the way cache invalidation is done, they leverage `AsyncLocalStorage`(ref https://nodejs.org/api/async_context.html#class-asynclocalstorage). 

There's a TC39 AsyncContext proposal (https://github.com/tc39/proposal-async-context) in Stage 2. Cloudflare added `async_hooks` to workerd (https://blog.cloudflare.com/workers-node-js-asynclocalstorage/)

vados-cosmonic:
Something that's worth looking at here (thanks @calvinrp): 

https://github.com/unjs/unenv

PR #234, opened 2025-03-26
feat: add node process.env builtin to access env vars

sd2k:
This uses wasi:cli/environment to get the environment variables provided by the host. The `process.env` object is a Proxy in this implementation so that we can call `get_environment` lazily, which enables wizer to run.

I needed this for a PoC so had Cursor help me out (it was surprisingly good) - it's probably not exactly the direction you want but I'm happy to iterate if it's in the right direction!

I think this relates to https://github.com/bytecodealliance/StarlingMonkey/issues/188?


squillace:
Hey, @andreiltd and I need this just like anyone will. Can't run in production without env vars for creds among other things and they can't be baked into the component, obvs. :-)

We'll see about having a look next week some time.

andreiltd:
Thank you for the patch! We had a brief discussion about it during the last JS SIG community meeting. While the functionality is definitely useful, adding just this small portion of the Node API as a built‑in doesn’t feel quite right at this stage.

A fully supported and currently working alternative for accessing environment variables with both `componentize-js` and `jco` is to use the WASI environment API. For example:

```js
import { getEnvironment } from 'wasi:cli/environment@0.2.3';

const entries = getEnvironment();
const env = Object.fromEntries(entries);

console.log(env.FOO);
```

Let us know if you'd like help adapting your use case to this approach.

PR #237, opened 2025-04-13
Fix parsing of the `--eval` CLI option

henrybw:
This patch fixes a bug in the CLI option parser that parses the argument to `--eval` twice, causing runtime initialization to interpret the argument as a script path, rather than an inline script. This is currently causing an example invocation in the README to fail:

```
wasmtime -S http starling.wasm -e "console.log('hello world')"

ReferenceError: Error loading module "console.log('hello world')"
(resolved path "console.log('hello world')"): No such file or directory
```

### DIFF
--- a/include/config-parser.h
+++ b/include/config-parser.h
@@ -77,8 +77,7 @@ public:
           config_->content_script_path.reset();
           i++;
         }
-      }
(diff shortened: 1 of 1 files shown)


tschneidereit (approved):
Ugh, thank you for fixing this, and apologies for such an embarrassing bug. I guess this really should be a `switch` ...

Issue #211, opened 2025-02-14
Optimize computing encoded `FormData` length


andreiltd:
As discussed here: https://github.com/bytecodealliance/StarlingMonkey/pull/200#discussion_r1952578511: It’s unfortunate that we’re only performing encoding to calculate the length of the encoded stream, after which the result is discarded. One possible solution would be to store the encoded output when constructing the `FormData`.

andreiltd:
@guybedford @tschneidereit maybe we can label this as a `good first issue`?

PR #241, opened 2025-04-19
Use JS module's length calculation for FormData values

FreeMasen:
This is an attempt to address #211 by using the [`JS::GetDeflatedUTF8StringLength`](https://searchfox.org/mozilla-central/source/js/src/vm/CharacterEncoding.cpp#71) to avoid encoding while calculating the length of values.

This change doesn't follow the suggestion to use `variant<JS::Heap<JS::Value>, HostString>` but I believe this would be equivalent to the existing implementation. If it would be preferable to use the memoization approach suggested I would be happy to close this and re-open when I can figure out how to implement that version.

## Testing

I've added a new test to the `integration` tests explicitly to target length calculation in this path, please feel free to request more test cases.


andreiltd (changes requested):
Hey, thank for looking into this!

I had to refresh my memory on this one. Generally speaking I don't think `Variant<JS::Heap<JS::Value>, HostString>` will work here because we would need both. Otherwise when calling `FormData#get` or similar, for the variant holding `HostString` we would need to call `core::decode` to get the `Value` from string. 

Overall I see few issues with the code in this patch:

- each call to `JS_GetStringCharAt` is internally calling `JS_EnsureLinearString` which potentially allocates and is very slow for iterating over characters, instead we should get linear string once and operate on its memory directly,
- `JS_EnsureLinearString` is fallible and its return value should be checked against `nullptr`
- `compute_unencoded_normalized_len` function should itself be fallible (because of potential allocation failures) so it should return `mozilla::Result`.

e.g:

```cpp
  JSString *linear = JS_EnsureLinearString(cx, value);
  if (!linear) {
    return mozilla::Err;
  }

  JS::AutoCheckCannotGC nogc;
  const char16_t *chars;
  size_t length;
  if (!JS_GetTwoByteLinearStringCharsAndLength(cx, linear, &chars, &length)) {
    return mozilla::Err;
  }

  size_t defalted_len = JS::GetDeflatedUTF8StringLength(linear);
  for (size_t i = 0; i < length; ++i) {
    auto c = chars[i];
    ...
  }
```

I appreciate the tests, although in this case I think the WPT test suite coverage is really good for `FormData` and the added test feel redundant. You can run the WPT tests like this:

```
just wpt-setup
just wpt-test
```

You should see the number of failures if your function is incorrect.

FreeMasen:
Thank you very much for the feedback. I believe I have the changes locally working but will likely not be able to clean them up to push until the evening US Central Time today or tomorrow.

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -42,6 +43,33 @@ size_t compute_normalized_len(std::string_view src) {
   return len;
 }
 
+std::optional<size_t> compute_unencoded_normalized_len(JSContext *cx, JS::HandleString value) {
```
FreeMasen, line 46: I ran into an issue resolving with the `mozilla::Result` type at this level and instead am using `std::nullopt` as an OOM indicator. If that isn't the right choice I can move this function into the namespace where OutOfMemory is defined.

builtins/web/form-data/form-data-encoder.cpp
```diff
@@ -42,6 +43,33 @@ size_t compute_normalized_len(std::string_view src) {
...
+  auto linear = JS_EnsureLinearString(cx, value);
+  if (!linear) {
+      return std::nullopt;
+  }
+  auto len = JS::GetDeflatedUTF8StringLength(linear);
+  size_t two_byte_length = JS::GetLinearStringLength(linear);
+  for (size_t i = 0; i < two_byte_length; i++) {
+    char16_t ch = JS::GetLinearStringCharAt(linear, i);
```
FreeMasen, line 54: I couldn't find a function named `JS_GetTwoByteLinearStringCharsAndLength`, the closest I could find was [`JS_GetTwoByteStringCharsAndLength`](https://searchfox.org/mozilla-central/source/js/src/jsapi.cpp#3323) but that requires that the string does not include latin1 characters. 

This seems to correctly return the `char16_t` regardless of the underlying encoding. 

I did find [`GetTwoByteLinearStringChars`](https://searchfox.org/mozilla-central/source/js/public/String.h#426) after I got this working and I can adjust this PR to use that if requested but I think that also asserts that the underlying characters are not latin1

andreiltd: Oh right, that's annoying. I think we have to use `JS::LinearStringHasLatin1Chars` and use either `TwoByte` or `Latin1` functions based on result.

andreiltd: Right, `JS::GetLinearStringCharAt` is already much better than original `GetStringCharAt`. Arguably, using the former still requires checking for each character in the loop whether it’s Latin or TwoByte. Clearly, you can move these checks out of the loop by performing them in advance. LLVM might be smart enough to optimize the checks, but that's just a guess without measuring it. :slightly_smiling_face: 

https://searchfox.org/mozilla-central/source/js/public/String.h#440

andreiltd: On the other hand we already make a lot checks in the loop so it might be just fine to have one more, if it makes the code more readable :)

andreiltd: One way of doing it "cleanly" would be through lambda with auto arguments or template function, perhaps?

Ignore if you don't like it:
```
std::optional<size_t> compute_unencoded_normalized_len(JSContext *cx, JS::HandleString str) {
  JSLinearString* lstr = JS_EnsureLinearString(cx, str);
  if (!lstr) {
    return std::nullopt;
  }

  auto normalized_len = [](const auto* chars, const auto len) {
    size_t extra = 0;
    for (size_t i = 0; i < len; i++) {
      auto ch = chars[i];
     
     // do stuff...
    
    return extra;
  };

  JS::AutoCheckCannotGC nogc;
  size_t extra = 0;

  const size_t len = JS::GetLinearStringLength(lstr);

  if (JS::StringHasLatin1Chars(str)) {
    const auto chars = GetLatin1LinearStringChars(nogc, lstr);
    MOZ_ASSERT(chars);
    extra = normalized_len(chars, len);
  } else {
    const auto chars = JS::GetTwoByteLinearStringChars(nogc, lstr);
    MOZ_ASSERT(chars);
    extra = normalized_len(chars, len);
  }

  return JS::GetDeflatedUTF8StringLength(lstr) + extra;
}
```

FreeMasen: Ok, I've pushed 2 commits, first with the lamba local to the new "unencoded" function. The second commit adds a template to allow using the same calculation for `std::string_view`. If it doesn't make sense to use the template just let me know and I can pop that one off before squashing

andreiltd (approved):
Thank you!

FreeMasen:
All squashed up and ready to merge, thank you for all your support and suggestions while I worked through this

PR #242, opened 2025-04-25
Add clang-tidy and clang-tidy-fix targets

andreiltd:
This enables clang based linting over c++ codebase. There are two new targets:

- `just lint` - runs clang-tidy on the codebase
- `just lint-fix` - runs clang-tidy with `-fix` option to automatically fix issues

The `.clang-tidy` file contains minimal configuration to get us started. See enabled `Checks`.

I was hoping to start the discussion on linting StarlingMonkey in general:
- do we want linting at all,
- should we try to apply the offered fixes,
- should we run the checks on CI,
- should we enforce the checks on PRs?

I hope we can review the errors from the linter and decide which are just the noise and which we want to fix.

cc: @tschneidereit


tschneidereit:
Some quick responses:

> * do we want linting at all,
Yes, I absolutely think we do
> * should we try to apply the offered fixes,
I haven't looked at them, but as long as none stick out as very annoying, I'm all for sticking with formatting defaults.
> * should we run the checks on CI,
Yes :)
> * should we enforce the checks on PRs?

Yes :)

The reason for the last two "yes"s is that that'll help avoid the situation where patches mix functional and formatting changes, making it much harder to identify the former.

andreiltd:
Thanks! OK, I will move this PR to draft then and try to come up with sensible lint defaults and fixes.

tschneidereit (approved):

tschneidereit:
This looks good to me, fwiw. Feel free to un-draft and land this once you feel comfortable with it

builtins/web/abort/abort-signal.cpp
```diff
@@ -259,9 +259,10 @@ bool AbortSignal::abort(JSContext *cx, HandleObject self, HandleValue reason) {
 bool AbortSignal::run_abort_steps(JSContext *cx, HandleObject self) {
   // To run the abort steps for an AbortSignal signal:
   // 1. For each algorithm of signal's abort algorithms: run algorithm.
-  auto algorithms = AbortSignal::algorithms(self);
+  auto *algorithms = AbortSignal::algorithms(self);
   for (auto &algorithm : *algorithms) {
-    if (!algorithm->run(cx)) return false;
+    if (!algorithm->run(cx)) { return false;
```
tschneidereit, line 264:
```suggestion
    if (!algorithm->run(cx)) {
      return false;
```

builtins/web/abort/weak-index-set.h
```diff
@@ -31,7 +30,7 @@ class WeakIndexSet {
   }
 
   WeakVec &items() { return items_;}
-  const WeakVec &items() const { return items_; }
+  [[nodiscard]] [[nodiscard]] [[nodiscard]] const WeakVec &items() const { return items_; }
```
tschneidereit, line 33: I hope we don't need to be quite as insistent here ;)
```suggestion
  [[nodiscard]] const WeakVec &items() const { return items_; }
```

builtins/web/abort/abort-controller.cpp
```diff
@@ -35,7 +35,7 @@ bool AbortController::abort(JSContext *cx, unsigned argc, JS::Value *vp) {
 
   RootedValue reason(cx, args.get(0));
   RootedObject signal(cx, JS::GetReservedSlot(self, Slots::Signal).toObjectOrNull());
-  if (!signal) {
+  if (signal == nullptr) {
```
tschneidereit, line 38: I'm not necessarily convinced that changing all the `[!]foo` checks to explicit comparisons increases clarity, personally. As such I'd mildly prefer we disable whatever check requires them. But at this point doing so would require a lot of work undoing all the changes, so I'll not insist on it.

builtins/web/blob.cpp
```diff
@@ -444,20 +442,19 @@ bool Blob::append_value(JSContext *cx, HandleObject self, HandleValue val) {
...
-      auto src = converted.data();
+      auto *src = converted.data();
       auto len = converted.length();
       return blob->append(src, src + len);
 
-    } else {
-      auto src = chars.ptr.get();
+    }       auto *src = chars.ptr.get();
       auto len = chars.len;
       return blob->append(src, src + len);
```
tschneidereit, line 451: (Alternatively, just fold this into a single line, inlining `src` and `len`)
```suggestion
    }
    auto *src = chars.ptr.get();
    auto len = chars.len;
    return blob->append(src, src + len);
```

builtins/web/abort/abort-controller.cpp
andreiltd: So this is the check:

https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html

I'm happy to revert the changes it shouldn't be too much work.

andreiltd: Sure! Makes sense. Let me try this out.

andreiltd:
>  It'd be great if you could look for those with a regex and fix them in one go.

Maybe it's worth, to just run `clang format` on the codebase as a very last commit in this PR?

andreiltd:
I kept my janitor hat on for a little bit longer and cleaned up ~clang diagnostics~ for invalid `offsetof` and disabled the warning globally with the last commit.

While disabling this warning is not ideal, starting from version 20.0, clang introduces the suppression maps option so we can selectively disable the warning only in problematic header files:

https://clang.llvm.org/docs/WarningSuppressionMappings.html

If it's OK, I will add an issue to remove the `-Wno-invalid-offsetof` compile flag and replace it with suppression mapping when we switch to the next `wasi-sdk`. WDYT?

tschneidereit:
I wasn't sure whether you were done iterating on it. Will take a look now.

tschneidereit:
> If it's OK, I will add an issue to remove the `-Wno-invalid-offsetof` compile flag and replace it with suppression mapping when we switch to the next `wasi-sdk`. WDYT?

Ah, I had been wondering why it was okay to disable these checks. That all makes sense to me, and as long as you/we don't forget to apply that change once we can update SDKs, I'm okay with this plan. Getting rid of all the pragmas sure is nice!

PR #248, opened 2025-06-26
VS Code debugger

itowlson:
Supersedes #219, on which it is based with a few tweaks / features / bug fixes (with @tschneidereit's involvement and consent).


debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
```diff
@@ -0,0 +1,677 @@
...
+    const componentRuntimeArgs = Array.from(config.componentRuntime.options).map(opt => {
+      return opt
+        .replace("${workspaceFolder}", workspaceFolder)
+        .replace("${component}", component);
+    });
+    componentRuntimeArgs.push(config.componentRuntime.envOption);
+    componentRuntimeArgs.push(
+      `STARLINGMONKEY_CONFIG="${config.jsRuntimeOptions.join(" ")}"`
+    );
+    componentRuntimeArgs.push(config.componentRuntime.envOption);
```
andreiltd, line 161: I think there's an important limitation when building runtime arguments this way: they won't work properly if you switch from using `wasmtime serve` to `wasmtime run` or `wasmtime run --invoke` (which we want to support for debugging scripts that don't target wasi/http).

Here's what happens: when we use `wasmtime run <module>`, the command parser treats everything that comes after the module name as input arguments for the WASM module itself, rather than as `wasmtime` options.

For example:

```
wasmtime run -S cli,tcp,inherit-network,http --invoke 'hello-user("Gordon Shumway")' out.wasm --env STARLINGMONKEY_CONFIG="--verbose -d" --env DEBUGGER_PORT=54954 
```

In this case, StarlingMonkey won't receive any of the environment variables because `wasmtime` interprets them as arguments to pass to the WASM module, not as configuration for the runtime itself.

tschneidereit:
@andreiltd thank you for your kind words!

Since Ivan and I both did meaningful parts of the work here, it'd be fantastic if you could do a proper review of the whole PR. I'll then do the work to address whatever feedback you'll have, and we can get this landed.

debugger/vscode-dap-extension/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+**`ts` workspace:** `npm run compile`
+
+**`spin-ts` workspace:** `spin build`
+
+(you will need to `npm install -g @bytecodealliance/componentize-js`)
+
+**MOAR DEETS**
+
+The content needs to be invoked via the `wasi:http/incoming-handler@0.2.*/handle` function. In other words, it needs to target the `wasi:http/proxy@0.2` world.
```
andreiltd, line 25: I'd hope that with the recent changes to `componentize-js` and StarlingMonkey this is no longer needed:
- https://github.com/bytecodealliance/StarlingMonkey/pull/249
- https://github.com/bytecodealliance/ComponentizeJS/pull/237

debugger/vscode-dap-extension/src/starlingMonkeyDebugger.ts
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  protected disconnectRequest(
+    _response: DebugProtocol.DisconnectResponse,
+    args: DebugProtocol.DisconnectArguments,
+    _request?: DebugProtocol.Request
+  ): void {
+    console.log(
+      `disconnectRequest suspend: ${args.suspendDebuggee}, terminate: ${args.terminateDebuggee}`
+    );
```
andreiltd, line 157: Should this send the response as well?

debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
```diff
@@ -0,0 +1,677 @@
...
+    this._sourceFile = this.normalizePath(program);
+
+    this.runDebugLoop();
+  }
+
+  async runDebugLoop() {
+    // We have a slightly odd mix here of state machine stuff (unsolicited messages)
+    // and request-response stuff.  Feels like there should be a better way.
+
+    while (true) {
```
andreiltd, line 274: It's unclear how we exit this loop. Should this break on some `message.type`?

debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
```diff
@@ -0,0 +1,677 @@
...
+        const parsed = JSON.parse(message);
+        resetMessageState();
+        this._messageReceived.resolve(parsed);
+      } catch (e) {
+        console.warn(`Illformed message received. Error: ${e}, message: ${partialMessage}`);
+        resetMessageState();
+      }
+    };
+
+    this._server = Net.createServer(socket => {
```
andreiltd, line 432: Should we check if server is already running before we start a new one?

debugger/vscode-dap-extension/src/starlingMonkeyDebugger.ts
itowlson: There's nothing to do: the JS DAP library equips us with a "default" response (line 151), and there's no need for us to modify anything about it (https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect, final paragraph).  And the DAP library takes care of the actual sending.

debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
itowlson: `"end"` is a JS event, not a message; it's handled at `starlingMonkeyDebugger.ts` line 123.  (The events and messages thing is a bit fiddly, especially coming in cold, because there's a lot of overlap.  I tried to tease them apart by declaring the RuntimeEventMap (events) and the RuntimeToInstanceMessage/InstanceToRuntimeMessage (socket messages) but it still doesn't jump out to the eye I realise.)

itowlson:
> IIUC, the debugging server in its current form does not support multiple sessions. Is that correct?

Yes, that's a limitation of the current implementation (I think @tschneidereit had plans for multi-component debugging, which we know we'll need, but wanted to get something out there in the meantime).  I'll look at detecting and erroring.

debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
tschneidereit: @itowlson I was wrong in what I just told you about this, unfortunately: we should indeed check whether the server is already running here. And in fact I found that I had a change to that effect, and to generally improve robustness around the communication, sitting around on my machine for a while now. I've now [pushed](https://github.com/tschneidereit/StarlingMonkey/commit/3b7954fe2133557e4b43a191daccc11e8e45ab20) this to my fork: it might be worth looking at, and perhaps even applying to whatever degree possible. My apologies for not sharing this earlier.

tschneidereit: No no, there's nothing you're misunderstanding: I should've put off replying here until the morning so as to be more awake. Instead, I confused myself into misunderstanding this.

Turns out the explanation of what's going on here I gave you yesterday was correct after all.

@andreiltd the way this works is that when a debug session is started in the extension, a dedicated server is started for just that session. Which means we should not check if another server is already running. If it is, then it's part of another concurrent debugging session. (I think we have other issues blocking those right now, but that's the theory at least.)

There is a global singleton server, but that one is started from under `ComponentRuntimeInstance.start`, and indeed guarded so it won't be started multiple times.

All of this could most decidedly be structured much better so we all wouldn't get as confused :/

andreiltd: Thanks for explanation!

debugger/vscode-dap-extension/src/starlingMonkeyRuntime.ts
tschneidereit: Yeah, I think that's the right thing to do for now. We can iterate in-tree as needed.

debugger/vscode-dap-extension/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Testing
+
+Use the `tests` directory (and add to it!).  Build commands:
+
+**`js` workspace:** `componentize-js --wit world.wit -o main.wasm --runtime-args "--enable-script-debugging" main.js`
+
+**`ts` workspace:** `npm run compile`
+
+**`spin-ts` workspace:** `spin build`
```
tschneidereit, line 76: This (and the folder it refers to) would be good to remove before landing. I think that really boils down to removing this line and the folder, but there might be other things I'm missing.

tschneidereit (approved):
One small tweak remaining, then we should go ahead and land this! 🥳

debugger/vscode-dap-extension/README.md
itowlson: @tschneidereit Where would you suggest we keep tests for non-Wasmtime hosts?  I think there's value in such tests existing.  For myself, I'd find it useful for them to be present in this repo, even if not everybody is able to run all of them, but happy to move them off to the other hosts' host orgs if preferred.  (Although if we did that it would still be useful to link them from here.)

tschneidereit: Such tests should indeed exist, yes! Given that we'll want to do integration with the Spin SDK more generally, I think the tests should logically live there, though. Additionally, hosting downstream embedding-specific tests here would open a can of worms, where it's not clear which ones we'd accept here and which we wouldn't.

PR #251, opened 2025-07-14
feat(crypto): add support for PKCS#8 key import

andreiltd:
- Implement PKCS#8 key import for ECDSA and RSA algorithms
- Add RAII wrappers for OpenSSL resources
- Remove deprecated openssl API calls


andreiltd:
Thank you for the review and I apologize for combining all the changes into a single PR.

Issue #95, opened 2024-07-29
Add an automated release process


tschneidereit:
Currently, StarlingMonkey doesn't contain an automated process to release binaries. This is something we should change, such that every merged PR results in a release including binaries for all of the project's build configurations.

PR #252, opened 2025-07-22
feat(ci): add release github workflow

andreiltd:
Closes: https://github.com/bytecodealliance/StarlingMonkey/issues/95

Release-As: 0.2.0


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,68 @@
...
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
```
vados-cosmonic, line 11: NIT: but it's safer to actually give *no* permissions at the top, and then granular permissions per step for each relevant job.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install Just
+        uses: taiki-e/install-action@just
+
+      - name: Install Rust 1.80.0
+        run: |
+          rustup toolchain install 1.80.0
```
vados-cosmonic, line 38: Probably a good idea to move this rust version up to a `matrix` value earlier rather than later!

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,68 @@
+name: release-please
+
+on:
+  push:
+    branches:
+      - main
```
vados-cosmonic, line 6: Is there any facility for making pre-releases? If not that's fine too -- but I've found it pretty useful

vados-cosmonic (approved):
LGTM :rocket: 

Left some nits but in general release-please is a great choice

.github/workflows/release-please.yml
vados-cosmonic: Ah OK not much we can do there then!

andreiltd:
We should probably start using https://www.conventionalcommits.org/en/v1.0.0/ from now on :)

andreiltd:
Hey @tschneidereit would you prefer landing this before or after ff140 update?

tschneidereit:
@andreiltd I agree that landing this before makes sense, so feel free to do that at any point.

andreiltd:
D'oh! This failed on main on creating a pull request despite the permissions for a job are explicitly granted:

```yaml
    permissions:
      contents: write
      issues: write
      pull-requests: write
```

@vados-cosmonic do you know if this also requires allowing jobs to create a pull request in the repository settings? How jco and componentize-js are configured?

vados-cosmonic:
I think what you're missing there is repository configuration! I think in your case it it may be this:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository

An image of what that section looks like (under `https://github.com/<org>/<repo>/settings/actions`):

image

We use a bot user in Jco, but I actually don't think Release Please requires that -- See [their instructions](https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials) and this sentence in particular:

> If your repository is in an organization, you may need to [permit github actions to create an approve PRs](https://stackoverflow.com/questions/72376229).

If that's already set for StarlingMonkey, then *maybe* you might need a bot user/explicitly permissioned token!

PR #253, opened 2025-07-22
fix(ci): give deploy job write permissions

andreiltd:
I'm not sure why I don't need this to deploy gh-pages on my fork, but hopefully this will fix the CI:

https://github.com/bytecodealliance/StarlingMonkey/actions/runs/16442845527/job/46467489433

Another possible issue might be that the workflow permissions for repo (settings > actions > general) are set to read only. I don't have access to verify this.


tschneidereit (approved):
You're correct: the repo is configured to only give actions read access by default. If I understand the documentation correctly, this change should indeed fix things, so let's hope it does!

PR #255, opened 2025-07-31
Update to SpiderMonkey 140

tschneidereit:
This is a big change in that it doesn't just do a SpiderMonkey update, but also changes how SpiderMonkey is integrated in the first place:

Instead of having the external [spidermonkey-wasi-embedding](https://github.com/bytecodealliance/spidermonkey-wasi-embedding) repo doing debug/release builds of SpiderMonkey as a static library, and StarlingMonkey just pulling those in, with this PR the build is integrated into StarlingMonkey's CMake setup.

The build system is still prepared to make use of pre-generated binaries, but right now nothing produces and publishes those. I'll follow this up with more commits adding both caching for CI and publishing of the binaries. Once that's done, hopefully the setup introduced here will just work with them. If not, I'll tweak things accordingly.


cmake/wasi-sdk.cmake
```diff
@@ -1,6 +1,6 @@
-set(WASI_SDK_VERSION 20 CACHE STRING "Version of wasi-sdk to use")
+set(WASI_SDK_VERSION 25 CACHE STRING "Version of wasi-sdk to use" FORCE)
```
andreiltd, line 1: Is there any reason why we don't go for the latest, which I belive is 27?

cmake/spidermonkey.cmake
```diff
@@ -21,24 +19,95 @@ endif()
...
+    set(SM_INCLUDE_DIR ${SM_LIB_DIR}/include)
+
+    file(GLOB SM_OBJS ${SM_LIB_DIR}/lib/*.o)
+
+    add_library(spidermonkey STATIC)
+    target_sources(spidermonkey PRIVATE ${SM_OBJS} ${CMAKE_CURRENT_BINARY_DIR}/null.cpp)
+    target_include_directories(spidermonkey PUBLIC ${SM_INCLUDE_DIR})
+    target_link_libraries(spidermonkey PUBLIC ${SM_LIB_DIR}/lib/libjs_static.a)
+else()
+    CPMAddPackage(NAME gecko-source
```
andreiltd, line 56: Does this support shallow cloning? I think external project has something like `GIT_SHALLOW 1`.

scripts/build-spidermonkey.sh
```diff
@@ -0,0 +1,118 @@
...
+weval=""
+if [[ $# -gt 1 ]] && [[ "$2" == "weval" ]]; then
+    weval=-weval
+fi
+mozconfig="$(pwd)/mozconfig-${mode}${weval}"
+objdir="obj-$mode${weval}"
+outdir="$mode${weval}"
+
+cat << EOF > "$mozconfig"
+ac_add_options --enable-project=js
```
andreiltd, line 50: Should we enable `lto`?  `--enable-lto=<full | thin>`

andreiltd (approved):
I've had some questions posted above, but none of these are blocking. LGTM! 🚀

scripts/build-spidermonkey.sh
tschneidereit: Good call: I did that. Makes linking take a lot longer, but that seems good, if anything :)

cmake/spidermonkey.cmake
tschneidereit: Shallow cloning was supposed to be done automatically. Turns out that doesn't actually work in `ExternalProject` itself, so I changed things pretty substantially to make shallow cloning an actual thing.

tschneidereit:
Oh right, sorry: I had addressed them, but never replied.

tschneidereit:
I'll hold off on merging this until we have the release out, as agreed.

PR #256, opened 2025-08-03
Try fixing release-please permissions

tschneidereit:
Let's see if this fixes the release permissions. It at least matches the [release-please action's documentation](https://github.com/googleapis/release-please-action#basic-configuration).

In looking at this, I also noticed that the published artifacts aren't quite what I expected—my apologies for not catching this earlier! There's a key difference between the `starling-raw.wasm` and `starling.wasm` builds, and we shouldn't rename the former to the latter for the released artifacts. Instead, I simply added both, so that they can both be made use of directly.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -5,13 +5,14 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
+  issues: write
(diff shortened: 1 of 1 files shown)


andreiltd (approved):
LGTM!

PR #257, opened 2025-08-03
chore(build): Move WPT support build-time to runtime config

tschneidereit:
Instead of requiring a CMake build config option to enable WPT support, with this change that's moved to the new `--wpt-mode` runtime flag.Instead of requiring a CMake build config option to enable WPT support, with this change that's moved to the new `--wpt-mode` runtime flag.

Also changes CI to only do one build of the runtime itself for all types of tests.


andreiltd:
Oh that's right I missed that, sorry!

Issue #123, opened 2024-08-27
SubtleCrypto Algorithms


guybedford:
If I could lobby for a few more algorithms to be supported 🙏 

- `RSA-PSS`
- `ECDSA` (at least P-256, but ideally all three curves)
- `Ed25519` from https://wicg.github.io/webcrypto-secure-curves/
- almost forgot, the goto JWS algorithm - `HMAC`

Also `pkcs8` and `spki` key format import/export are not to be overlooked. Developers LOOOVE to use in favour of JWKs.

I have a comprehensive [test suite](https://github.com/panva/jose#supported-runtimes) for different runtimes' webcrypto implementations that I could unload on js-compute-runtime eventually.

_Originally posted by @panva in https://github.com/fastly/js-compute-runtime/issues/440#issuecomment-1535339296_

cupofjoakim:
More context of why it would be great to have this. If you use session tokens from google identity platform/firebase and want to verify them, the public keys are only available as x509's, requiring SPKI support: https://www.googleapis.com/identitytoolkit/v3/relyingparty/publicKeys

https://firebase.google.com/docs/auth/admin/manage-cookies#verify_session_cookies_using_a_third-party_jwt_library

PR #277, opened 2025-10-01
feat(crypto): Add SPKI format support for ECDSA and RSA keys

cupofjoakim:
In an attempt to work on #123 I wanted to try to add SPKI format support.

Please review thoroughly as I have _no_ experience in cpp or in crypto and my confidence level is not great. This should be seen as an attempt, not necessarily a complete solution. There's a solid risk that I've missed something that is required - I've basically gone for the SPKI todo's in the crypto-algorithm.cpp file.
  
  Changes:
  - Implement SPKI import for ECDSA (P-256, P-384, P-521)
  - Implement SPKI import for RSA-PKCS1-v1_5
  - Add integration tests for SPKI key import
  - Enforce verify-only usage for SPKI public keys

### DIFF
--- a/tests/wpt-harness/expectations/WebCryptoAPI/import_export/ec_importKey.https.any.js.json
+++ b/tests/wpt-harness/expectations/WebCryptoAPI/import_export/ec_importKey.https.any.js.json
@@ -78,10 +78,10 @@
     "status": "FAIL"
   },
   "Good parameters: P-256 bits (spki, buffer(91), {name: ECDSA, namedCurve: P-256}, false, [])": {
-    "status": "PASS"
+    "status": "FAIL"
   },
   "Good parameters: P-256 bits (spki, buffer(59, compressed), {name: ECDSA, namedCurve: P-256}, false, [])": {
-    "status": "PASS"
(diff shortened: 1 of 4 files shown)


andreiltd:
Hey, thank you for this! I think for CI to pass you will have to verify/update WPT tests. You can run the tests locally:

```
just wpt-setup
just wpt-test
```

you can update expectations with:

```
just wpt-update
```

The missing tests are certainly something you want to verify as this may indicate hitting some assertions or crashes.

cupofjoakim:
@andreiltd Hey, I updated the tests and I'm unsure if everything is right. It did add a ton of passes, but it had some other changes I was not prepared for, and I'd like to check with you if it's as expected before committing fully.

You can see the results here: https://github.com/cupofjoakim/StarlingMonkey/pull/1/files

What i'm worried about:
- Two seemingly unrelated files have been deleted
- A lot of tests have been renamed (removing substring `importVectorKeys step: `)

Don't want to pollute the PR unnecessarily

andreiltd (approved):
This looks great, thank you!

PR #305, opened 2026-03-04
chore(deps): bump weval

andreiltd:
Use the latest release of weval (0.4.1) that includes improvement for memory packing. The componentize.sh script produces 30% smaller binary with new weval.


tschneidereit (approved):
Very nice!

Issue #311, opened 2026-06-08
body.blob() does not set Blob.type from Content-Type header


godronus:
When calling `.blob()` on a `Request` or `Response`, the resulting `Blob` always has an empty `type`, regardless of the `Content-Type` header on the request/response.

Per the [Fetch spec - consume body](https://fetch.spec.whatwg.org/#dom-body-blob), the Blob's type should be set to the body's extracted MIME type — derived from the `Content-Type` header, ASCII-lowercased.
  
 ### WPT coverage
  
 `fetch/api/body/mime-type.any.js` has 20 tests covering this behaviour. 
The file currently cannot run: it contains a top-level `bodyContainer.clone()` call that throws `TypeError` because `Response.clone()` is not yet implemented, which aborts the file before any tests are registered. Once `Response.clone()` is implemented, 14 of those 20 tests will fail due to this bug.
  
### Root cause
  
In `builtins/web/fetch/request-response.cpp`, `parse_body<BodyReadResult::Blob>` hardcodes `contentType` to the empty string:
  
```cpp
} else if constexpr (result_type == RequestOrResponse::BodyReadResult::Blob) {
    JS::RootedString contentType(cx, JS_GetEmptyString(cx));  // always empty
    JS::RootedObject blob(cx, blob::Blob::create(cx, std::move(buf), len, contentType));
```

Fix

The FormData branch immediately below already does the correct thing — reads the Content-Type header via Headers::lookup and calls extract_mime_type. The Blob branch needs the same treatment, with the result ASCII-lowercased before being passed to Blob::create. The absent-header case (where blob.type should remain "") is already handled correctly by the current fallback.

PR #312, opened 2026-06-08
feat: implement Response.clone()

### LINKED ISSUE #184: Stream disturbed test failures
Since we usually try and have tracking issues for WPT failures, this is to track the new failures in  the recently added tests in https://github.com/bytecodealliance/StarlingMonkey/blob/main/tests/wpt-harness/expectations/fetch/api/request/request-disturbed.any.js.json.

godronus:
## Implement Response.clone() (WHATWG Fetch)

Implements `Response.clone()` per the WHATWG Fetch spec. All 21 WPT tests in
`fetch/api/response/response-clone.any.js` pass.

## Approach

Response bodies are modelled as **default** (non-byte) `ReadableStream`s, whose tee shares a single
chunk object between both branches. Fetch's "clone a body" requires the clone to receive independent
copies, so `Response::clone` implements the WHATWG `ReadableStreamDefaultTee` algorithm with
`cloneForBranch2 = true`:

- A single reader drains the source body stream.
- Each chunk is enqueued **unchanged** into branch1 (the original response's body), preserving
  object identity, and a `JSAutoStructuredCloneBuffer` copy is enqueued into branch2 (the clone's
  body).
- The copy is made **eagerly**, in the source read, before either consumer observes the chunk — a
  lazy copy would be read-order dependent and would duplicate already-mutated bytes if the original
  branch is read and mutated in place first.

Cancellation is tracked per branch (`canceled1`/`canceled2`): a cancelled branch is dropped from the
read loop while the other continues.

`sc_callbacks` is not needed — body chunks are `TypedArray`/`ArrayBuffer`/`DataView`, handled by the
built-in structured clone.

## Design notes

- **Cost:** one structured-clone copy per chunk, clone branch only; branch1 is zero-copy. Same copy
  count as the spec byte-stream tee. `highWaterMark` is 0 and chunks are pulled on demand, so the
  body is never fully buffered.
- **Byte streams:** modelling bodies as byte streams would let the native byte-stream tee provide
  this directly (`CloneAsUint8Array`, no hand-rolled tee), but that is a large change across all
  body I/O and is out of scope here.
- **`Request::clone`** currently tees without copying branch2; aligning it is a reasonable
  follow-up.

## Changes

`builtins/web/fetch/request-response.h`

- Added `Response::clone` declaration.

`builtins/web/fetch/request-response.cpp`

- Added `#include "js/StructuredClone.h"`.
- `Request::bodyUsed_get` / `Response::bodyUsed_get`: fall back to `JS::ReadableStreamIsDisturbed`
  when the `BodyUsed` slot is false, so `bodyUsed` reports `true` after `body.cancel()` and other
  external stream disturbances.
- Added the tee implementation: `CloneTeeSlot` (shared-state layout), `clone_tee_pump`,
  `clone_tee_read_then_handler`, `clone_tee_read_catch_handler`, `clone_tee_cancel_algorithm`,
  `clone_tee_pull_algorithm`.
- Implemented `Response::clone`: acquires a default reader on the body, creates two
  `NativeStreamSource` branches over shared tee state, and assigns branch1 to the original response
  and branch2 to the clone.
- Registered `JS_FN("clone", Response::clone, 0, JSPROP_ENUMERATE)`.

`tests/wpt-harness/expectations/fetch/api/response/response-clone.any.js.json`

- New file — 21/21 passing.

## WPT results

`fetch/api/response/response-clone.any.js` — 21/21 pass.

Side-effect fixes from the `ReadableStreamIsDisturbed` fallback in `bodyUsed_get`:

- `fetch/api/response/response-stream-disturbed-6.any.js` — 5 FAIL → PASS (partially resolves #184)
- `fetch/api/response/response-stream-disturbed-by-pipe.any.js` — 2 FAIL → PASS
- `fetch/api/request/request-disturbed.any.js` — 1 FAIL → PASS
- `fetch/api/abort/general.any.js` — 2 FAIL → PASS

`fetch/api/body/mime-type.any.js` — new expectations file (6 PASS, 14 FAIL). The file previously
aborted during initialisation on a top-level `clone()` that threw `TypeError`; it now loads, and the
14 failures expose a pre-existing gap where `parse_body<Blob>` does not populate `blob.type` from
the `Content-Type` header. Tracked in #311.

### Downstream testing: PASSING

https://github.com/G-Core/FastEdge-sdk-js/blob/main/integration-tests/test-application/handlers/response-clone.ts

### DIFF
--- a/builtins/web/fetch/request-response.cpp
+++ b/builtins/web/fetch/request-response.cpp
@@ -1197,7 +1197,14 @@ bool RequestOrResponse::maybe_stream_body(JSContext *cx, JS::HandleObject body_o
   // First, handle direct forwarding of incoming bodies.
   // Those can be handled by direct use of async tasks and the host API, without needing
   // to use JS streams at all.
-  if (is_incoming(body_owner)) {
+  //
+  // This fast path is only valid while the body has *not* been reified into a JS
+  // ReadableStream. Once a `BodyStream` exists, JS is the authoritative source of the
+  // body bytes and must be honored. In particular `clone()` tees the body and installs
+  // a tee branch into this object's `BodyStream`; if we forwarded the raw incoming
+  // handle here we would bypass that branch and double-read the single host body
+  // (the other tee branch reads the same handle), corrupting both consumers.
+  if (is_incoming(body_owner) && !body_stream(body_owner)) {
     auto *source_body = incoming_body_handle(body_owner);
     auto *dest_body = destination->body().unwrap();
     auto res = dest_body->append(ENGINE, source_body, finish_outgoing_body_streaming, nullptr);
--- a/builtins/web/fetch/request-response.cpp
+++ b/builtins/web/fetch/request-response.cpp
@@ -23,6 +23,7 @@
 #include "js/Conversions.h"
 #include "js/JSON.h"
 #include "js/Stream.h"
+#include "js/StructuredClone.h"
 #include "js/experimental/TypedData.h"
 #include "mozilla/ResultVariant.h"
 
@@ -1362,7 +1363,19 @@ bool Request::body_get(JSContext *cx, unsigned argc, JS::Value *vp) {
 
 bool Request::bodyUsed_get(JSContext *cx, unsigned argc, JS::Value *vp) {
   METHOD_HEADER(0)
-  args.rval().setBoolean(RequestOrResponse::body_used(self));
+  bool used = RequestOrResponse::body_used(self);
+  if (!used) {
+    JSObject *stream = RequestOrResponse::body_stream(self);
+    if (stream) {
+      JS::RootedObject body(cx, stream);
+      bool disturbed = false;
+      if (!JS::ReadableStreamIsDisturbed(cx, body, &disturbed)) {
+        return false;
+      }
+      used = disturbed;
+    }
+  }
+  args.rval().setBoolean(used);
   return true;
 }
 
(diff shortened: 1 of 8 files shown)


andreiltd:
Thank you! Looks good. It's a bit unfortunate that we cannot use SpiderMonkey's `ReadableStreamTee` here directly.

That said, I think those changes will break `RequestOrResponse::maybe_stream_body` assumptions: https://github.com/bytecodealliance/StarlingMonkey/blob/main/builtins/web/fetch/request-response.cpp#L1223

IIUC `Response::clone()` can install tee branches into `BodyStream`, but leaves the original incoming response handle in place and `maydbe_body_stream` will bypass the tee branch?

See the comment:

```
 // Those can be handled by direct use of async tasks and the host API, without needing
 // to use JS streams at all.  
```

Also, this has to be rebased to include: https://github.com/bytecodealliance/StarlingMonkey/pull/313, because it will change wpt expectations. Is that correct?

andreiltd:
@tschneidereit did you want to have another look? If I'm not mistaken you're fresh out of the stream implementation 🙂

godronus:
> That said, I think those changes will break `RequestOrResponse::maybe_stream_body`

Good catch — fixed in the latest push.

One-liner in maybe_stream_body: also gate the incoming fast-path on `!body_stream(body_owner)`, so once a `BodyStream` exists (`.body` access or a `clone()` tee) we fall through to the JS-stream path and honor the tee branch instead of draining the raw handle. Sits in the shared path, so `Request::clone` gets the same fix for free.

Impact: basically nil for normal apps — proxy / header-rewrite / fetch-and-return never reify `BodyStream`, so they keep the zero-copy host append. It only diverts for cloned incoming responses, which then stream through JS in 8 KB chunks rather than a single host append — cost proportional to body size, and those callers were already paying that on the branch they read. Correctness over a micro-opt on the one path that needs it.

I did look at gating on an explicit "was teed" flag to keep the fast-path for the bare .body-reify case, but that only helps "reify .body, never read it, return the same object" — which doesn't really happen — so I went with the simpler guard.

All WPT + unit tests still green.


> Also, this has to be rebased to include: https://github.com/bytecodealliance/StarlingMonkey/pull/313, because it will change wpt expectations. Is that correct?

yes, WPT expectations will have some "FAIL" with this branch alone and rebasing #313 will make them all "PASS"
**REBASED**

andreiltd (approved):
Thanks!

PR #313, opened 2026-06-08
fix(fetch): body.blob() sets Blob.type from Content-Type header

### LINKED ISSUE #311: body.blob() does not set Blob.type from Content-Type header

godronus:
## Summary

Fixes #311.

When calling body.blob() on a Request or Response, the returned Blob was always created with an empty type, regardless of the Content-Type header on the body. This was uncovered by the WPT pass improvements in #312.

Per the Fetch spec 5.2:

▎ If type is non-null, then set object's type to the result of parse a MIME type with type's value.

This fix reads the Content-Type header from the owning request/response when constructing the Blob in parse_body, extracts the MIME type using the existing extract_mime_type helper, and passes the normalized type through Blob::normalize_type before creating the Blob object.

Changes

- builtins/web/blob.h / blob.cpp: Promotes normalize_type from a file-local free function to a public static method on Blob, so request-response.cpp can call it without duplication.
- builtins/web/fetch/request-response.cpp: In parse_body, after constructing the body buffer, looks up the Content-Type header on the owning object, parses the MIME type, normalizes it, and passes it when creating the Blob.
- tests/wpt-harness/expectations/fetch/api/body/formdata.any.js.json: Updates one previously-failing WPT test to PASS (Consume multipart/form-data headers case-insensitively) — the test was failing because it asserted on the Blob.type value derived from the Content-Type header.

Testing

The fix is covered by the WPT test suite. The Consume multipart/form-data headers case-insensitively test in `fetch/api/body/formdata.any.js` now passes.

Running WPT test suite after merging [Response.clone (#312)](https://github.com/bytecodealliance/StarlingMonkey/pull/312) creates a new `fetch/api/body/mime-type.any.js` expectations file.  (20 PASS, 0 FAIL)

### DIFF
--- a/builtins/web/blob.cpp
+++ b/builtins/web/blob.cpp
@@ -26,60 +26,6 @@ template <typename T> bool validate_type(T *chars, size_t strlen) {
   return true;
 }
 
-// 1. If type contains any characters outside the range U+0020 to U+007E, then set t to the empty string.
-// 2. Convert every character in type to ASCII lowercase.
-JSString *normalize_type(JSContext *cx, HandleValue value) {
-  JS::RootedString value_str(cx);
-
-  if (value.isObject() || value.isString()) {
-    value_str = JS::ToString(cx, value);
-    if (!value_str) {
-      return nullptr;
-    }
-  } else if (value.isNull()) {
-    return JS::ToString(cx, value);
-  } else {
-    return JS_GetEmptyString(cx);
-  }
-
-  auto *str = JS::StringToLinearString(cx, value_str);
-  if (!str) {
-    return nullptr;
-  }
-
-  std::string normalized;
-  auto strlen = JS::GetLinearStringLength(str);
-
-  if (strlen == 0U) {
-    return JS_GetEmptyString(cx);
-  }
-
-  if (JS::LinearStringHasLatin1Chars(str)) {
-    JS::AutoCheckCannotGC nogc(cx);
-    const auto *chars = JS::GetLatin1LinearStringChars(nogc, str);
-    if (!validate_type(chars, strlen)) {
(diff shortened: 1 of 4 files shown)


andreiltd (approved):
LGTM, thank you!

PR #317, opened 2026-06-27
fix(form-data): preserve leading BOM in multipart text parts

watilde:
Per the Fetch spec, a multipart/form-data part without a filename is decoded with "UTF-8 decode without BOM", which preserves a leading U+FEFF. The parser used encoding_new_decoder_with_bom_removal, which strips it, so "\uFEFFhello" was returned as "hello".

Use encoding_new_decoder_without_bom_handling, matching the spec text already quoted in the surrounding comment. Add an integration test.


andreiltd (approved):
Thank you!

PR #318, opened 2026-06-27
fix(console): console.assert() logs on failure, not on success

watilde:
console.assert()'s condition was inverted: `if (!condition) return` made it stay silent when the assertion failed and print "Assertion failed" when it passed — the opposite of the cited Console spec, whose step 1 is "If condition is true, return."

Also use JS::ToBoolean() instead of Value::toBoolean(): the latter asserts the argument is already a boolean and crashes a debug build (UB in release) on a non-boolean condition such as console.assert(0). The spec coerces condition to a boolean.

Add an e2e test asserting the serve stdout for truthy/falsy and boolean/non-boolean conditions.


andreiltd (approved):
Makes sense, thank you for the fix.

PR #319, opened 2026-06-27
fix(structured-clone): preserve Blob.type across structuredClone

watilde:
structuredClone(blob) serialized only the byte data and reconstructed the Blob with an empty type, so structuredClone(new Blob(['x'], {type: 'text/plain'})).type returned "" instead of "text/plain". A Blob's type is part of its serialized state per the File API.

Append the type as a length-prefixed UTF-8 string after the byte data on write, and read it back on read. Add an integration test covering a typed Blob, an empty type, and content/type round-tripping.

### DIFF
--- a/builtins/web/structured-clone.cpp
+++ b/builtins/web/structured-clone.cpp
@@ -1,5 +1,7 @@
 #include "structured-clone.h"
 #include "blob.h"
+#include "decode.h"
+#include "encode.h"
 
 #include "dom-exception.h"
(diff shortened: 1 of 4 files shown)


andreiltd (changes requested):
Thanks! Could you update WPT expectations? I think the easiest way to do that is to run:

```
just wpt-setup
just wpt-update
```

This will update new passes. Also I believe the tests added here largely duplicates WPT coverage so we can drop them.

PR #320, opened 2026-06-27
fix(url): validate name in URLSearchParams.append()

watilde:
URLSearchParams.append() never checked whether stringifying the name argument succeeded. For a non-stringifiable name (a Symbol, or an object whose toString throws), core::encode_spec_string returns a SpecString with a null data pointer and a pending exception, which append handed straight to the Rust params_append — crashing the process in slice::from_raw_parts (UB in release) instead of throwing.

Null-check the encoded name and propagate the pending exception, exactly as delete()/has()/get()/set() already do. Add an integration test.

### DIFF
--- a/builtins/web/text-codec/text-decoder.cpp
+++ b/builtins/web/text-codec/text-decoder.cpp
@@ -39,7 +39,9 @@ bool TextDecoder::decode(JSContext *cx, unsigned argc, JS::Value *vp) {
   }
 
   bool stream = false;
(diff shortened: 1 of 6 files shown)


andreiltd (approved):
LGTM!

andreiltd:
@watilde could you resolve conflicts? It is good to go otherwise.

watilde:
@andreiltd Thanks for reviewing. Rebased onto main

